PR #1, opened 2015-11-09
Typo in README

johnnypeck:
Change "are" to "as."

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # KnpUOAuth2ClientBundle
 
 This bundle integrates with [league/oauth2-client](http://oauth2-client.thephpleague.com/)
-and allows you to easily create different "providers" are services in your container.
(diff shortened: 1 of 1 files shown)


weaverryan:
Awesome, thanks :)

weaverryan:
And btw, this isn't quite useable yet, but will be in an alpha, useable state pretty soon :)

okwinza:
@weaverryan dont wanna open up a new issue for this so i just wanted to let u know about the most annoying bug in HWIOauthBundle in case you've missed it: hwi/HWIOAuthBundle#149

It gave me a lot of headache back then so would be nice to avoid it in this bundle.
Cheers!

weaverryan:
@okwinza ah, thanks - I *am* actually going to open an issue so that I don't forget about that :).

See #2

PR #3, opened 2015-11-12
Google provider config

outspaced:


weaverryan:
@outspaced Sorry for the delay on this! I needed to get the bundle to its proper state. Could you rebase your pull request to the new system for adding providers? I think you'll like it - it's a whole heck of a lot cleaner.

Thanks!

outspaced:
@weaverryan I have rebased the code, but most of the additions from this PR has now been covered by a different PR that was merged.  The only thing remaining is a unit test for the Google provider/client.

weaverryan:
@outspaced you're right - I'm sorry about that! In my zest to merge providers, I completely didn't realize we had a duplicate (the Google provider).

In this case, we don't need the test - the test above this (for Facebook) tests the system that the Google one plugs into (we *did* need this test before when each provider has more responsibility, but it's all centralized now).

Thanks for taking the time to do this and for finding my other (critical) docs mistake :).

PR #5, opened 2016-01-20
Update README.md

arturolinares:
Fixed a typo that can cause the login redirection to fail


weaverryan:
Ah, thank you! That's very important - I have made that mistake, and the error is difficult to figure out :).

Issue #7, opened 2016-01-25
We need to add a lot more providers!


weaverryan:
We should support as many providers as possible, and there are _many_: https://github.com/thephpleague/oauth2-client/blob/master/README.PROVIDERS.md
- [x] Google #3
- [x] Instagram
- [x] LinkedIn #9
- [ ] VK #10 (was reverted)

... and I would accept any other additions

javiereguiluz:
Wouldn't Google, Facebook, Twitter, LinkedIn, Instagram and GitHub cover 99% of the needs? I'd say those are enough for the first version of the bundle.

Babacooll:
I can manage the Twitch provider if nobody has planed to do it.

fatihkahveci:
If we can add Yahoo and Yandex provider too it will be cool for first release.

weaverryan:
@javiereguiluz I agree, I won't wait for providers to make a release. I'm not afraid of making many more releases after as we add more providers.

@Babacooll I've never used Twitch, but that would be rad!

kbond:
How could Twitter be added? It is my understanding it is [oauth1](https://github.com/thephpleague/oauth1-client).

weaverryan:
Ah yea, we can't support Twitter. I'm not opposed to supporting OAuth1 - obviously Twitter makes a good case for it, but I don't know if it's the *only* one. I also don't know if it'll be a huge pain to try to integrate it in cleanly (i.e. hacky), or if it could be done cleanly. There *is* an OAuth1 library from the PHP league... and perhaps we should *not* support it, but mention something in the README (e.g. use this OAuth1 library directly if you need something like Twitter). Afterall, this bundle is mostly a think integration layer.

jmolivas:
@kbond @weaverryan I found a project for Twitter https://github.com/Maslosoft/oauth2-twitter but is not working properly.

pmaselkowski:
Hello, 
Maslosoft/oauth2-twitter project is on hold until twitter fully implements oauth2.

weaverryan:
fwiw - here's a quick answer for others that find this on how you could use the phpleage oauth1 library in Symfony to connect with Twitter: http://knpuniversity.com/blog/KnpUOAuth2ClientBundle#comment-3086507464

proArtex:
Vote for twitter provider

jokoast:
Isn't Twitter using Oauth2 ? Why isn't it implemented ?

[...]

PR #9, opened 2016-01-25
LinkedIn Provider Added

fatihkahveci:
for #7 

Also if we can create new wiki i can added linkedin usage sample too.


weaverryan:
This looks awesome - thanks @fatihkahveci!

About the usage, I'll create a tutorial for the library soon - because I agree that for beginners (especially with OAuth), it's not very understandable until you can see how all the code should flow. I haven't used LinkedIn's API before, is there anything special to its flow that should be highlighted?

fatihkahveci:
@weaverryan almost same like other providers. Also tutorial will be awesome for usage.

PR #12, opened 2016-01-29
Google Provider Added

fatihkahveci:
for #7

Google Provider added also i change duplicate connectAction method.

### DIFF
--- a/DependencyInjection/Providers/GoogleProviderConfigurator.php
+++ b/DependencyInjection/Providers/GoogleProviderConfigurator.php
@@ -7,8 +7,15 @@ use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 class GoogleProviderConfigurator implements ProviderConfiguratorInterface
 {
     public function buildConfiguration(NodeBuilder $node)
-    {
-        // no custom options
+    {    
+       $node
+            ->scalarNode('access_type')
+                ->defaultValue('online')
+            ->end()
+            ->scalarNode('hosted_domain')
+                ->defaultValue('')
+            ->end()
+        ;
     }
 
     public function getProviderClass()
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -123,7 +123,7 @@ class FacebookController extends Controller
      *
      * @Route("/connect/facebook/check", name="connect_facebook_check")
      */
-    public function connectAction(Request $request)
+    public function connectCheckAction(Request $request)
```
weaverryan: Ah, good catch - thank you :)

DependencyInjection/Providers/GoogleProviderConfigurator.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+
+class GoogleProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        // no custom options
```
weaverryan: What about `hostedDomain` - I see that here: https://github.com/thephpleague/oauth2-google. It looks like there's also an `accessType`: https://github.com/thephpleague/oauth2-google/blob/master/src/Provider/Google.php#L20. I'm not familiar with either of these - I just spotted them on the driver :)

fatihkahveci: @weaverryan Right hostedDomain and accsesType just optional parameters. But we must add this your right my bad :+1:

weaverryan:
Merged! Thanks so much! Btw, if you want to add others (which would be awesome), you don't need to update the README - it updates itself via my super awesome (super messy) script: https://github.com/knpuniversity/oauth2-client-bundle/blob/master/bin/update_readme

Thanks!

fatihkahveci:
@weaverryan This reminde me https://github.com/NARKOZ/hacker-scripts that script :D.

PR #15, opened 2016-02-04
Corrected documentation to use clients, not providers in configuration

outspaced:


bocharsky-bw:
:+1: according to the https://github.com/knpuniversity/oauth2-client-bundle/blob/master/DependencyInjection/Configuration.php#L17

weaverryan:
Ah, thanks! This is why I have the rest of the configuration generated automatically :)

PR #20, opened 2016-02-25
Fix scope usage

blazarecki:
Hi,

This PR allow to use scope.

For now, `scopes` was send but it should be `scope`

I test it for github.
https://developer.github.com/v3/oauth/#response

Here documentation from other provider:

- https://developers.facebook.com/docs/reference/javascript/FB.login/v2.5
- https://developer.linkedin.com/docs/oauth2

What do you think ?

### DIFF
--- a/Client/OAuth2Client.php
+++ b/Client/OAuth2Client.php
@@ -45,7 +45,7 @@ class OAuth2Client
     {
         $options = array();
         if (!empty($scopes)) {
-            $options['scopes'] = $scopes;
+            $options['scope'] = $scopes;
         }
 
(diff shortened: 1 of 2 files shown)


bocharsky-bw:
:+1: Really, it should be `scope` not `scopes`. Google also uses a `scope` parameter name: https://developers.google.com/identity/protocols/OAuth2UserAgent

weaverryan:
Awesome, thank you @blazarecki! I didn't notice this because the OAuth servers I tested with were defaulting to the default scopes. Thanks for catching this!

PR #29, opened 2016-06-12
Implemented user_fields into the google provider so that a user can request more than the defined default fields of a g-user.

MM-Psiiirus:


MM-Psiiirus:
Any concerns??

weaverryan:
Nope, this makes perfect sense to me! New release just tagged.

Thanks Toni!

PR #31, opened 2016-07-24
Require php version

bocharsky-bw:


weaverryan:
I think this is quite sensible :)

PR #32, opened 2016-07-24
Use '^' instead of '~' for required package versions

bocharsky-bw:
> The ^ operator behaves very similarly but it sticks closer to semantic versioning, and will always allow non-breaking updates.


weaverryan:
:thumbsup: Thanks!

PR #33, opened 2016-07-24
Add Instagram provider

bocharsky-bw:
In addition to the #7

### DIFF
--- a/Client/Provider/InstagramClient.php
+++ b/Client/Provider/InstagramClient.php
@@ -4,7 +4,6 @@ namespace KnpU\OAuth2ClientBundle\Client\Provider;
 
 use KnpU\OAuth2ClientBundle\Client\OAuth2Client;
 use League\OAuth2\Client\Provider\InstagramResourceOwner;
-use League\OAuth2\Client\Provider\ResourceOwnerInterface;
 use League\OAuth2\Client\Token\AccessToken;
 
 class InstagramClient extends OAuth2Client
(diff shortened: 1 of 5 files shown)


Client/Provider/InstagramClient.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param AccessToken $accessToken
+     * @return InstagramResourceOwner
+     */
+    public function fetchUserFromToken(AccessToken $accessToken)
+    {
+        return parent::fetchUserFromToken($accessToken);
+    }
+
+    /**
+     * @return InstagramResourceOwner|ResourceOwnerInterface
```
weaverryan: You can remove the `| ResourceOwnerInterface`, since InstagramResourceOwner implements this

weaverryan:
I think this looks good to me :). Is this still WIP?

Btw, I can do it, but if you want, you can execute `php bin/update_readme` to auto-update the README with the documentation for this provider.

Thanks! This is a great provider to have!

bocharsky-bw:
Yes, it's still WIP, I've not tested it yet.

BTW, I've tried to execute `bin/update_readme `, but it doesn't update README for me. I'll look closer on it later, maybe I missed something.

weaverryan:
Ah! That's because I think we forgot one spot: https://github.com/knpuniversity/oauth2-client-bundle/blob/master/DependencyInjection/KnpUOAuth2ClientExtension.php#L31

But yea - once you test it, it will catch any last things we're missing :)

Thanks!

bocharsky-bw:
Done!

weaverryan:
Awesome! Thanks @bocharsky-bw!

PR #34, opened 2016-08-18
Dev dependencies should include the compatible PHPunit version

curry684:
Added the correct `phpunit` and `php-code-coverage` dependencies so everyone can run the tests right out of the box after a fork.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,9 @@
     },
     "require-dev": {
         "league/oauth2-facebook": "^1.1",
-        "symfony/security-guard": "^2.8|^3.0"
+        "symfony/security-guard": "^2.8|^3.0",
+        "phpunit/phpunit": "^5.5",
+        "phpunit/php-code-coverage": "^4.0"
     },
     "suggest": {
         "symfony/security-guard": "For integration with Symfony's Guard Security layer"
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,8 @@
     "require-dev": {
         "league/oauth2-facebook": "^1.1",
         "symfony/security-guard": "^2.8|^3.0",
-        "phpunit/phpunit": "^5.5",
-        "phpunit/php-code-coverage": "^4.0"
+        "phpunit/phpunit": "^4.8",
+        "phpunit/php-code-coverage": "^2.2"
     },
     "suggest": {
         "symfony/security-guard": "For integration with Symfony's Guard Security layer"
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": " >=5.5.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -24,7 +24,9 @@
     },
     "require-dev": {
         "league/oauth2-facebook": "^1.1",
-        "symfony/security-guard": "^2.8|^3.0"
+        "symfony/security-guard": "^2.8|^3.0",
+        "phpunit/phpunit": "^4.8",
```
bocharsky-bw: I think we should allow 5.x version of PHPUnit in this case too with `^4.8|^5.0`, right?

curry684: It would also need finetuning of the `php-code-coverage` since they're intertwined, but yeah it could work. In general though I myself favor just picking a single major to avoid having differences between developers. I would just postpone the 5.x upgrade until PHP 5.5 is dropped in this project itself.

weaverryan: I'm fine with using just one major, as this - because being a `dev` dependency, this won't cause any problems with end-user projects, correct (i.e. an end-user could still install `phpunit/phpunit` version 5 in their project, without this conflicting)?

And, is there a precedence for including `php-code-coverage` as a dependency in other OS projects? I'm just not familiar with this - and mostly, I want to follow what's most expected/standard.

Thanks!

curry684: I recommend sticking to the single major for the universally reproducible results. As dev requirements are installed per project this means it's fine if you have 3 projects on the same machine depending on PHPunit `^4.6`, `5.1.*` and `^5.2` - each will get their own copy.

As dev requirements themselves are never propagated to external projects you can in practice do whatever the hell you want with them, it's your project  😄   `php-code-coverage` itself has been installed [25.5 million times](https://packagist.org/packages/phpunit/php-code-coverage) so I'd think it enough precedent   😉 

I'll pick this up along with #36, do expect a freakish PR coming up from that with all files modified hehe.

bocharsky-bw: My question is why do we really need `phpunit/php-code-coverage` in this project? I agree with PHPUnit, we use it for tests, but we don't use `php-code-coverage` here. As for me adding a dependency which we don't use in a project just in case is a bad idea which leads to overhead and even could confuse other devs.

curry684:
I agree but it broke when I removed it hehe. I'll investigate why it did that.

curry684:
Apparently that fixed itself along the way.

composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": " >=5.5.0",
+        "php": " >=5.5.9",
```
bocharsky-bw: 👍

weaverryan:
Awesome, thanks!

Issue #36, opened 2016-08-19
Add a coding style checker and enforce it


curry684:
Adding a `.php_cs` file would make the project more consistent internally, I noticed a lot of missing comments, long array notation et al.

(I'll be happy to do this if asked but it's commonly considered maintainer's task and privilege to set and enforce standards).

weaverryan:
Hi @curry684!

I would warmly welcome this, and would love if you would add it - I don't often use a cs fixer (and so don't know much about the `.php_cs` files), but it does make a lot of sense for OS projects, and I love the automation of such things. We should probably just base our file off of Symfony's https://github.com/symfony/symfony/blob/master/.php_cs

Cheers!

curry684:
The Symfony one is actually far too specialized and custom given their use of generated code et al, I'll whip up a simple one from one of my other projects.

curry684:
Closed by #37

PR #37, opened 2016-08-19
Coding style enforced

curry684:
As discussed in #36


Exception/InvalidStateException.php
```diff
@@ -1,11 +0,0 @@
-<?php
```
weaverryan: Could we keep everything at the root - so not in `src/`? That would keep it consistent, at least, with other bundles.

.gitattributes
```diff
@@ -0,0 +1,28 @@
...
+*.gz binary
+*.zip binary
+*.7z binary
+*.ttf binary
+*.eot binary
+*.woff binary
+*.pyc binary
+*.pdf binary
+*.woff binary
+*.woff2 binary
```
weaverryan: What's this file based off of? It seems like it's got a lot of unnecessary details

Exception/InvalidStateException.php
weaverryan: Hmm, actually, let's keep them in `src/` - it's indeed a bit more organized.

weaverryan:
Hey @curry684!

I asked just one question about the gitattributes. But this is really really clean - thanks a lot! I'll merge as soon as we resolve that last item.

Cheers!

weaverryan:
And, I want to tag a new release - I'll do it after this.

curry684:
I chose to reorganize the folders because having a bin and tests folder makes it messy. Without those I agree it wouldn't have real added value, now it isolates the code namespace to its own folder.

As for the gitattributes it's a lesson hard learnt on my end after once wasting 2 hours trying to find a file corruption issue to find out it was git itself mistaking a PNG for plaintext and converting 'newlines' before push. Since then I always have it in my project templates, even if actually irrelevant like here, just to avoid running into the same issue later on hehe. I have no problem removing it, it doesn't have real added value right now nor will it likely have that one day for this bundle.

weaverryan:
Thanks @curry684! And I've just tagged a new release :)

PR #38, opened 2016-08-21
More code cleaned up, test coverage increased

curry684:
More generic finetuning and best practices, added some more phpDoc and some tests to increase line coverage to 85%. Made some private functionality that could limit extension protected.


src/Client/OAuth2Client.php
```diff
@@ -136,9 +145,17 @@ public function getOAuth2Provider()
...
+    public function isStateless()
+    {
+        return $this->isStateless;
+    }
+
+    /**
      * @return \Symfony\Component\HttpFoundation\Request
      */
-    private function getCurrentRequest()
+    protected function getCurrentRequest()
```
weaverryan: I actually disagree with this: we should only add extension points when people ask and give us a good reason - that'll help maintenance in the future. If we keep this private, the can still fetch the Request if they want to, but injecting the `request_stack` service into directly into some base class.

src/Client/OAuth2Client.php
```diff
@@ -136,9 +145,17 @@ public function getOAuth2Provider()
     }
 
     /**
+     * @return bool
+     */
+    public function isStateless()
+    {
+        return $this->isStateless;
+    }
```
weaverryan: Same here - this is an unnecessary extension point, until someone gives us a good reason :)

src/Client/ClientRegistry.php
```diff
@@ -10,17 +10,26 @@
...
 
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class ClientRegistry
+class ClientRegistry implements ContainerAwareInterface
 {
-    private $container;
+    use ContainerAwareTrait;
```
weaverryan: This also changes `$container` from private to protected, which I don't want... until someone has a use-case :)

curry684:
I don't agree with your comments on `protected` variables and functions from a software architect's point of view. In a properly set up OOP library actually nearly *everything* should be `protected` instead of `private`, because you the developer can never foresee what other people ever want to do with your library.

Take the `$container` change as a practical use case: exactly why would you want to keep it private? Does the class' other functionality inherently break if a derived class accesses the container? Or even if he overwrites it? In my opinion it doesn't, see there is no *reason* to forbid reading and writing it. Private visibility should be reserved for properties that will fatally break if 'abused', like the primary key of a Doctrine entity. In nearly all other cases you're only making decisions for others that they can make fine themselves.

Just my 5 cents, we can agree to disagree and I'll revert those changes.

weaverryan:
@curry684 In Symfony's core, we take the exact opposite stance :). But, it's definitely subjective.

Here's an old blog post about that decision: http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html. In short, private makes it much easier for us to make changes, without breaking end-user code. But if there's a use-case for protected, and we decide it's the right extension point to add, we should do that.

Cheers!

curry684:
Yeah I know the post, I just tend to be on the pragmatist end of things  :wink: 

I'll revert the protected changes tonight, code is on my laptop at home.

curry684:
Done  👍

weaverryan:
Awesome, thank you!

PR #40, opened 2016-11-08
Changed provider link as current one returned 404 error

maciekpaprocki:


bocharsky-bw:
I don't much like refer to the list of files, but looks like it's the best option here 👍

bocharsky-bw:
But what do you think about referring to the https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/league.md instead? It has one more link to the third-party providers inside

maciekpaprocki:
Links on this page to third party don't work either, but probably good
shout.

I sent pull request to them too.

Feel like an asshole for doing only commits with link change and getting
everyone notified.

Thanks for quick feedback.
Maciej

Best regards
Maciej Paprocki

mobile: 07884432073
email: maciekpaprocki@gmail.com
www: paprocki.co.uk

bocharsky-bw:
@maciekpaprocki Yep, it's a bit confusing when links change :/

Btw, your PR will be declined. The reason is that broken links on GitHub are working for their website: http://oauth2-client.thephpleague.com/ . So we should use links to this website instead of GitHub. And yes, I know, they have not a perfect solution for building links on this website, but we can't do anything with it. So let's use this one:
http://oauth2-client.thephpleague.com/providers/league/

maciekpaprocki:
Yes, I agree, I just sent an update. 

That's weird that they have such a problem. Surely it's better to look on it on their docs page either way. 

Thanks

bocharsky-bw:
👍

bocharsky-bw:
Thank you @maciekpaprocki for the discussion here and this useful fix!

PR #45, opened 2017-01-07
Bitbucket Client added

luchianenco:
OAuth Bitbucket Client added, tests added

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "symfony/dependency-injection": "^2.8|^3.0",
         "symfony/routing": "^2.7|^3.0",
         "symfony/http-foundation": "^2.7|^3.0",
-        "league/oauth2-client": "^1.0"
+        "league/oauth2-client": "^1.0",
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -17,7 +17,8 @@
         "symfony/dependency-injection": "^2.8|^3.0",
         "symfony/routing": "^2.7|^3.0",
         "symfony/http-foundation": "^2.7|^3.0",
-        "league/oauth2-client": "^1.0"
+        "league/oauth2-client": "^1.0",
+        "stevenmaguire/oauth2-bitbucket": "^0.1.2"
```
bocharsky-bw: It's specific to the Bitbucket provider, we don't need to depend on it globally. If users want this feature - they will install it manually. See other providers

luchianenco: ok I will remove it from global require

weaverryan:
Thanks @luchianenco! Well-done :). I've just released 1.6.0 with this.

Cheers!

luchianenco:
welcome ;)

PR #47, opened 2017-01-09
added Odnoklassniki, Slack, Vimeo, Yahoo and Yandex providers

luchianenco:
my opinion is that config.yml code becomes too long for readme file. It would be great to create a separate file with all configuration options


weaverryan:
Wow, this AMAZING! You've basically doubled our providers in 2 days - really cool :). I'll tag a new release right now (I don't usually like to wait - I want people to be able to use the new stuff).

Btw, about the README, 2 things:

1) You're right that the config section is long... but I don't think it's a problem (it's at the bottom). Also, it's nice for people to be able to just search (via their browser - e.g. cmd+f) for something and have the README all on one page.

2) I should have told you earlier, but I didn't know you were going to send so many providers! But, there is an automated script to build the README config from the configuration files - `bin/update_readme`. You can run that to automatically update the README.

Thanks!

PR #49, opened 2017-01-26
Allow installing the new OAuth client 2.x in dependencies

### LINKED ISSUE #48: OAuth2 client libraries like Facebook needs "league/oauth2-client": "^2.0" to be installed!
I would like to switch to this library from HWIOAuthBundle (we like so much Guard 👍), but client libraries like Facebook, Github, Instagram, Google, Linkedin, Bitbucket (and maybe others...) have a required dependency with `"league/oauth2-client": "^2.0"` that's locked at the `"league/oauth2-client": "^1.0"` version in this bundle. How we can resolve the issue?

bocharsky-bw:
Fix #48

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "psr-4": { "KnpU\\OAuth2ClientBundle\\Tests\\": "tests/" }
     },
     "require-dev": {
-        "league/oauth2-facebook": "^1.1",
+        "league/oauth2-facebook": "^1.1|^2.0",
         "symfony/security-guard": "^2.8|^3.0",
         "phpunit/phpunit": "^4.8"
     },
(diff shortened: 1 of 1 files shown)


weaverryan:
Um... so the tests do pass... but all the runs in the test matrix use some version of 1.x of the library...  (you can see it in the `composer update` line in travis). I'm not sure exactly why... I would expect it to grab 2.x on a fresh composer update...

bocharsky-bw:
Yeah, I see... I'm trying to install dependencies locally to see if it grab new version - maybe it's just due to the Travis cache

bocharsky-bw:
The new version installation is suppressed by the `league/oauth2-facebook` dependency

bocharsky-bw:
Now the new versions 2.x were installed on Travis for OAuth client and facebook

weaverryan:
Of *course*!!! Brilliant! And tests still pass! Tagging a new version now!

PR #53, opened 2017-03-24
Fix optional configuration in Google provider

pyrech:
Avoid notice `Undefined index: access_type` on [this line](https://github.com/knpuniversity/oauth2-client-bundle/tree/ad82b0f780e1ad54dd7de2d353027d8384891080/src/DependencyInjection/Providers/GoogleProviderConfigurator.php#L44) when optional configuration is not given

### DIFF
--- a/src/DependencyInjection/Providers/GoogleProviderConfigurator.php
+++ b/src/DependencyInjection/Providers/GoogleProviderConfigurator.php
@@ -19,9 +19,11 @@ class GoogleProviderConfigurator implements ProviderConfiguratorInterface
         // todo - add the comments as help text, and render in README
         $node
             ->scalarNode('access_type')
+                ->defaultValue('')
                 ->info('Optional value for sending access_type parameter. More detail: https://developers.google.com/identity/protocols/OpenIDConnect#authenticationuriparameters')
             ->end()
             ->scalarNode('hosted_domain')
+                ->defaultValue('')
                 ->info('Optional value for sending hd parameter. More detail: https://developers.google.com/identity/protocols/OpenIDConnect#hd-param')
             ->end()
             ->arrayNode('user_fields')


bocharsky-bw:
Perfect sense 👍 

@weaverryan Btw, probably we should use Symfony OptionsResolver component here to make error messages more obvious?

pyrech:
Changed the PR to use add default values :+1:

src/DependencyInjection/Providers/GoogleProviderConfigurator.php
```diff
@@ -19,9 +19,11 @@ public function buildConfiguration(NodeBuilder $node)
         // todo - add the comments as help text, and render in README
         $node
             ->scalarNode('access_type')
+                ->defaultValue('')
                 ->info('Optional value for sending access_type parameter. More detail: https://developers.google.com/identity/protocols/OpenIDConnect#authenticationuriparameters')
             ->end()
             ->scalarNode('hosted_domain')
+                ->defaultValue('')
```
bocharsky-bw: Not sure, but maybe we need `null` instead for `access_type` and `hosted_domain` options?

src/DependencyInjection/Providers/GoogleProviderConfigurator.php
```diff
@@ -19,9 +19,11 @@ public function buildConfiguration(NodeBuilder $node)
...
                 ->info('Optional value for sending hd parameter. More detail: https://developers.google.com/identity/protocols/OpenIDConnect#hd-param')
             ->end()
             ->arrayNode('user_fields')
```
bocharsky-bw: We need a default value (an empty array) for the `user_fields` option too, right?

src/DependencyInjection/Providers/GoogleProviderConfigurator.php
pyrech: Hum, I don't know, I used the defaults proposed in the README [here](https://github.com/knpuniversity/oauth2-client-bundle#configuration).
Let me know what is you preference :wink:

src/DependencyInjection/Providers/GoogleProviderConfigurator.php
pyrech: I tested without defining default value for `user_fields` and it worked fine. So I suppose arrayNode / prototype make default value useless

weaverryan:
Thanks for reporting and fixing the bug! I added a bit more work on #54 - I would appreciate it if you tell me if it makes sense for you :). When I merge that, I'll tag a new release.

Thanks!

PR #54, opened 2017-03-30
[Google] Fixing undefined array indexes + extra request data

weaverryan:
Making sure that if config is *not* specified, that we don't include it in the request!

See #53


weaverryan:
Cool! Let's rock and roll - tagging a release right now...

PR #55, opened 2017-06-08
Update service definition for autowiring

brianfreytag:
As of Symfony 3.3, autowire_types is deprecated.

This adds an alias for the knpu.oauth2.registry service definition.

### DIFF
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -13,6 +13,9 @@
             <argument type="service" id="service_container" />
             <argument /> <!-- argument added dynamically -->
         </service>
+	
+	<!-- Add service alias for autowiring -->
+	<service id="KnpU\OAuth2ClientBundle\Client\ClientRegistry" alias="knpu.oauth2.registry" public="false" />
 
         <!-- Set an alias for easier use -->
         <service id="oauth2.registry" alias="knpu.oauth2.registry" />


brianfreytag:
When attempting to autowire the client registry:

```
azure.client_fetcher:
    class: AppBundle\Azure\Client\ClientFetcher
    autowire: true
```

```
use KnpU\OAuth2ClientBundle\Client\ClientRegistry;

class ClientFetcher
{
    /** @var ClientRegistry */
    private $clientRegistry;

    /**
     * @param ClientRegistry $clientRegistry
     */
    public function __construct(ClientRegistry $clientRegistry)
    {
        $this->clientRegistry = $clientRegistry;
    }
    ...
}
```

 I received:

```
Autowiring services based on the types they implement is deprecated since Symfony 3.3 
and won't be supported in version 4.0. You should rename (or alias) the "knpu.oauth2.registry" service 
to "KnpU\OAuth2ClientBundle\Client\ClientRegistry" instead.
```

I fixed this based on the upgrade path here: 

https://github.com/symfony/symfony/blob/master/UPGRADE-3.3.md#dependencyinjection

brianfreytag:
@weaverryan failure is false (hhvm)

src/Resources/config/services.xml
```diff
@@ -13,6 +13,9 @@
             <argument type="service" id="service_container" />
             <argument /> <!-- argument added dynamically -->
         </service>
+	
+	<!-- Add service alias for autowiring -->
+	<service id="KnpU\OAuth2ClientBundle\Client\ClientRegistry" alias="knpu.oauth2.registry" public="false" />
```
bocharsky-bw: Shouldn't we rename `knpu.oauth2.registry` to `KnpU\OAuth2ClientBundle\Client\ClientRegistry` and do the reverse?
```xml
<service id="knpu.oauth2.registry" alias="KnpU\OAuth2ClientBundle\Client\ClientRegistry" public="false" />
```

brianfreytag: Per the official 3.3 UPGRADE documentation, they recommend they do it the way I have it in the PR.

```
<service id="annotations.reader" class="Doctrine\Common\Annotations\AnnotationReader" public="false" />
<service id="Doctrine\Common\Annotations\Reader" alias="annotations.reader" public="false" />
```

weaverryan:
Thanks @brianfreytag! I've just tagged 1.10.0 :)

brianfreytag:
Thank you @weaverryan

PR #57, opened 2017-09-07
New Providers added

luchianenco:
I have added some new providers.


weaverryan:
Wow, this is amazing! That's a ton of work! Thanks for this - you just gave us MANY MANY new providers :).

Once I get the tests passing again, I'll tag a new version of the lib.

Cheers!

PR #61, opened 2017-09-21
Fixed example path check for getCredentials().

aurelijusrozenas:


bocharsky-bw:
Nice catch! We really have `/connect/facebook/check` in the README

Thanks!

PR #62, opened 2017-09-26
Set an alias to every client service

### LINKED ISSUE #60: Deprecation messages for created service
With the following basic config:
```
knpu_oauth2_client:
    clients:
        # the key "facebook" can be anything, it
        # will create a service: "knpu.oauth2.client.facebook"
        facebook:
            # this will be one of the supported types
            type: facebook
```
The following deprecation messages arise:
> Autowiring services based on the types they implement is deprecated since Symfony 3.3 and won't be supported in version 4.0. You should rename (or alias) the "knpu.oauth2.provider.facebook" service to "League\OAuth2\Client\Provider\Facebook" instead.

larzuk91:
Fix #60 

I feel like the _service Id_ and _alias name_ are inverted, but tests does not pass if I do it the other way around

### DIFF
--- a/src/DependencyInjection/KnpUOAuth2ClientExtension.php
+++ b/src/DependencyInjection/KnpUOAuth2ClientExtension.php
@@ -38,6 +38,7 @@ use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\Definition\Processor;
+use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
@@ -51,6 +52,9 @@ class KnpUOAuth2ClientExtension extends Extension
 
     /** @var array */
     private $configurators = [];
+
+    /** @var array */
+    private $duplicateProviderTypes = [];
 
     /** @var array */
     private static $supportedProviderTypes = [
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -214,6 +214,8 @@ private function configureProviderAndClient(ContainerBuilder $container, $provid
             $clientDefinition->addMethodCall('setAsStateless');
         }
 
+        $container->setAlias($clientClass, $clientServiceKey);
```
weaverryan: This looks great! Two minor things:

```php
$container->setAlias($clientClass, new Alias($clientServiceKey, false));
```

This will create a private alias... no reason for the user to need to fetch this service directly from the container with that id.

Second. There's one edge case: if someone (for some reason) configures *two* clients with the same provider (e.g. 2 facebook clients). In that case, the alias would be created twice, the second would override the first. It would be better to not create *any* alias in this case, so that the user receives an exception if they try to autowire it (it will be up to them to create an alias if they want to autowire).

How to do this easily? Let's add a new private property to the class: `private $duplicateProviderTypes = array()`. Then, we can do something like:

```php
// add an alias, but only if a provider type is used only 1 time
if (!in_array($providerType, $this->duplicateProviderTypes)) {
    // alias already exists? This is a duplicate type, record it
    if ($container->getAlias($clientClass)) {
        $this->duplicateProviderTypes[] = $providerType;
    } else {
        // all good, add the alias
        $container->setAlias($clientClass, new Alias($clientServiceKey, false));
    }
}
```

larzuk91:
Done!

src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -214,6 +218,17 @@ private function configureProviderAndClient(ContainerBuilder $container, $provid
             $clientDefinition->addMethodCall('setAsStateless');
         }
 
+        // add an alias, but only if a provider type is used only 1 time
+        if (!in_array($providerType, $this->duplicateProviderTypes)) {
+            // alias already exists? This is a duplicate type, record it
+            if ($container->getAlias($clientClass)) {
```
weaverryan: `hasAlias()` instead. My bad :)

weaverryan:
Boom!

PR #63, opened 2017-10-07
Amazon, Fitbit, Heroku, Paypal, Stripe Clients added

luchianenco:


weaverryan:
Thanks for this @luchianenco! Just give us a few days to review it - it's on our list :).

src/DependencyInjection/Providers/AmazonProviderConfigurator.php
```diff
@@ -0,0 +1,54 @@
...
+class AmazonProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        // no custom options
+    }
+
+    public function getProviderClass(array $config)
+    {
+        return 'Luchianenco\OAuth2\Client\Provider\Amazon';
```
larzuk91: I'm just wondering why not using the syntax `Amazon::class`?

larzuk91:
Only one minor comment, but everything looks great

Thanks @luchianenco!

src/DependencyInjection/Providers/AmazonProviderConfigurator.php
luchianenco: well, I tried to keep the style the repository had before ;)
of course it can be a next PR ;) to change in all Configurators

larzuk91: Oh, yeah, maybe @weaverryan can tell us the true reason behind this :)

bocharsky-bw: I think it's just legacy from the past, but since we require PHP `>=5.5.9` in `composer.json` now changing to `::class` seems reasonable to me, but yes, in another PR and for all occurrences 👍

src/DependencyInjection/Providers/PaypalProviderConfigurator.php
```diff
@@ -0,0 +1,60 @@
...
+namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+
+class PaypalProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        $node
+            ->scalarNode('is_sandbox')
```
bocharsky-bw: Looks like it should be `booleanNode()` instead, right?

README.md
```diff
@@ -609,6 +662,23 @@ knpu_oauth2_client:
...
+            # must be "paypal" - it activates that type!
+            type: paypal
+            # add and configure client_id and client_secret in parameters.yml
+            client_id: %paypal_client_id%
+            client_secret: %paypal_client_secret%
+            # a route name you'll create
+            redirect_route: connect_paypal_check
+            redirect_params: {}
+            # When true, client uses Paypal Sandbox urls.
+            # is_sandbox: ''
```
bocharsky-bw: This should be `false` as value by default instead of an empty string

src/DependencyInjection/Providers/AmazonProviderConfigurator.php
bocharsky-bw: Uh oh, probably I was wrong! The main problem is that we do not have strict dependencies on those classes, i.e. user installs them if needed. So, it'll throw an error if the class does not found. That's why here we use strings which postpone this check until actual call. Sorry for misleading you in my previous message 😇

README.md
luchianenco: `bin/update_readme` uses '' instead of boolean value. I will updated it manually but after next usage to update readme it will be again `is_sandbox: ''`

bocharsky-bw: Hm, I thought it should be `false` after `scalarNode()` -> `booleanNode()` tweak

luchianenco: I have updated update_readme to set boolean values...please check if it is fine

bin/update_readme
```diff
@@ -92,6 +93,11 @@ foreach (KnpUOAuth2ClientExtension::getAllSupportedTypes() as $type) {
                 // *should* ? come out looking like valid-ish YAML
                 ? json_encode($child->getDefaultValue())
                 : '{}';
+
+        } elseif ($child instanceof BooleanNode) {
+            $defaultValue = $child->getDefaultValue()
+                ? $child->getDefaultValue()
```
bocharsky-bw: Should be `true`

bocharsky-bw:
Thanks for the `bin/update_readme` tweak! Just one comment left.. and it'll look great for me

bin/update_readme
luchianenco: do u mean ; `$defaultValue = $child->getDefaultValue() ? 'true'`

bocharsky-bw: Not quite, I mean `? 'true'` on this line, i.e.:

```php
 $defaultValue = $child->getDefaultValue()
    ? 'true'
    : 'false';
```

luchianenco: that's what I meant ;) sorry just shorted the expression :)
ok I  will do

src/DependencyInjection/Providers/AmazonProviderConfigurator.php
larzuk91: Ahh, that makes sense, thanks for the clarification

weaverryan:
Woohoo! Sorry for the delay - but this is awesome! I'll tag a new release shortly :)

luchianenco:
@weaverryan I can understand :) too many events

weaverryan:
Haha, yea... a bit of traveling... and things are pretty busy right now with Symfony 4 prep. Busy... but exciting :) 

Cheers!

Issue #65, opened 2017-11-02
Symfony 4 support


slootjes:
Is this compatible with Symfony 4? If not, what work needs to be done to make it work? If yes, can it be added to composer.json so it will pass the compatiblity checks?

weaverryan:
Let's find out! :) #66

slootjes:
Great, I'll try it tomorrow and will post my findings.

weaverryan:
Just tagged 1.13.0 with Symfony 4 support - the tests all passed, so it should work fine. But if you find an issue, let us know!

slootjes:
Awesome, thanks for your fast response :)

slootjes:
I can not load 1.13 with composer in a v4.0.0-BETA2 project just yet:


Your requirements could not be resolved to an installable set of packages.

```
  Problem 1
    - Installation request for knpuniversity/oauth2-client-bundle ^1.13 -> satisfiable by knpuniversity/oauth2-client-bundle[1.13.0].
[...]

PR #66, opened 2017-11-02
Let's see if we support Symfony4!

weaverryan:
Go tests go! See #65


weaverryan:
Apparently we are already Symfony 4 compat :)

bocharsky-bw:
Now I see only Symfony 3.3 packages were installed, so Symfony 4 support wasn't even tested :/

PR #67, opened 2017-11-02
New clients added

luchianenco:
Another portion of clients: Salesforce, Zendesk, Strava, Uber, Unsplash, PSN, Mollie are added

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ via Composer:
 | [Stripe](https://github.com/adam-paterson/oauth2-stripe)              | composer require adam-paterson/oauth2-stripe        |
 | [Strava](https://github.com/Edwin-Luijten/oauth2-strava)              | composer require edwin-luijten/oauth2-strava        |
 | [Uber](https://github.com/stevenmaguire/oauth2-uber)                  | composer require stevenmaguire/oauth2-uber          |
+| [Unsplash](https://github.com/hughbertd/oauth2-unsplash)              | composer require hughbertd/oauth2-unsplash          |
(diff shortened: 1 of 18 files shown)


weaverryan:
Ahh... so cool! You rock @luchianenco! Check out our clients list now - https://github.com/knpuniversity/oauth2-client-bundle#step-1-download-the-client-library - 35! Woo!

luchianenco:
thanks ;) 
many providers are a bit outdated...do not support ResourceOwnerInterfacer

weaverryan:
Many in our list are outdated? Or many other client libraries are outdated (so we can't add them)?

luchianenco:
Many client libraries doesn't support ResourceOwnerInterface and therefore we cannot use them as I understand.

weaverryan:
Hmm, yea that makes sense -we require them to return that. Well, probably in those cases, those libraries aren't very used anyways :)

luchianenco:
Some of them, what I would like to add in the first row are Envato, MeetUp, Reddit, Twitch and some others that have really big usage...there are many also that we can add but as I understand they are not so popular.

weaverryan:
Hmm. Yea, I looked at a few of those. They're indeed implemented incorrectly - e.g. https://github.com/tpavlek/oauth2-twitch/blob/ead8e17d393610b35d06f6e42723a75590833a3c/src/Provider/Twitch.php#L130 should return a `ResourceOwnerInterface` (https://github.com/thephpleague/oauth2-client/blob/a887c33ff02ec7f92029903e9e7e0fc5a1c52583/src/Provider/AbstractProvider.php#L742), but it does not. 

That's a bug in their code. So, to support these, the libraries themselves would need to be improved, or replaced with new libs :/

PR #69, opened 2017-11-07
Ten more clients added 

luchianenco:
Let's increase the number to 45 ;)
Box, Buffer, CanvasLMS, Clever, DevianArt, Discord, Elance, Eventbrite, Foursquare and Mail.Ru are added

I use this list of providers [http://oauth2-client.thephpleague.com/providers/thirdparty/](url).
Maybe you have another one? Thx

### DIFF
--- a/src/DependencyInjection/Providers/CanvasLMSProviderConfigurator.php
+++ b/src/DependencyInjection/Providers/CanvasLMSProviderConfigurator.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * OAuth2 Client Bundle
+ * Copyright (c) KnpUniversity <http://knpuniversity.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+
+class CanvasLMSProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        $node
+            ->scalarNode('canvas_instance_url')
+                ->isRequired()
+                ->info('URL of Canvas Instance (e.g. https://canvas.instructure.com)')
+            ->end()
+            ->scalarNode('purpose')
+                ->defaultValue('')
+                ->info(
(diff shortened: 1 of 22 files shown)


src/DependencyInjection/Providers/CanvasLMSProviderConfigurator.php
```diff
@@ -0,0 +1,68 @@
...
+            ->scalarNode('canvas_instance_url')
+                ->isRequired()
+                ->info('URL of Canvas Instance (e.g. https://canvas.instructure.com)')
+            ->end()
+            ->scalarNode('purpose')
+                ->defaultValue('')
+                ->info(
+                    'This can be used to help the user identify which instance of an application this token is '
+                    .'for. For example, a mobile device application could provide the name of the device.'
+                )
```
weaverryan: Nice!

src/DependencyInjection/Providers/CanvasLMSProviderConfigurator.php
```diff
@@ -0,0 +1,68 @@
...
+        return 'smtech\OAuth2\Client\Provider\CanvasLMS';
+    }
+
+    public function getProviderOptions(array $config)
+    {
+        return [
+            'clientId' => $config['client_id'],
+            'clientSecret' => $config['client_secret'],
+            'canvasInstanceUrl' => $config['canvas_instance_url']
+        ];
```
weaverryan: I think the `purpose` needs to be mentioned here?

luchianenco: oh, forgot :(

luchianenco:
You are right!!! Doc generator is really nice tool ... thanks for it ;)

bocharsky-bw:
Oh my... where do you find them? I even have not heard about some of them :p Great job!

> You're making me so happy that we made the documentation auto-generate

Lol, that's true :)

luchianenco:
@bocharsky-bw your are right .... some of them are really unknown ... and I think sometimes : do we really need them all or not :)

bocharsky-bw:
Good question! IMO having them all is fine until their libraries are supported / not deprecated. I'd say it worth us nothing to have them all... well, with your BIG help, it worth us nothing ;)

luchianenco:
Good, I will add them all from time to time, and then I will try to update deprecated libraries that are a bit popular :)

weaverryan:
Sweeeeet!

PR #71, opened 2017-12-20
Keycloak provider: Set optional configuration items default values to null instead of an empty string to avoid a PHP warning

stephaneseng:
## Problem

Using [knpuniversity/oauth2-client-bundle (v1.15.0)](https://github.com/knpuniversity/oauth2-client-bundle/tree/1.15.0) with [stevenmaguire/oauth2-keycloak (v2.0.0)](https://github.com/stevenmaguire/oauth2-keycloak/tree/2.0.0), when a Keycloak provider is declared without an `encryption_key_path` the following PHP warning is thrown, even though this configuration item is optional:
`PHP Warning: file_get_contents(): Filename cannot be empty in [...]/vendor/stevenmaguire/oauth2-keycloak/src/Provider/Keycloak.php on line 242`.

## Explanations

`Stevenmaguire\OAuth2\Client\Provider\Keycloak` providers are instanciated by `KnpU\OAuth2ClientBundle\DependencyInjection\ProviderFactory` using an array of options generated by `KnpU\OAuth2ClientBundle\DependencyInjection\Providers\KeycloakProviderConfigurator`, using values read from the configuration items.

However, because the default value of `encryption_key_path` is an empty string (see [here](https://github.com/knpuniversity/oauth2-client-bundle/blob/1.15.0/src/DependencyInjection/Providers/KeycloakProviderConfigurator.php#L33)) and because `Stevenmaguire\OAuth2\Client\Provider\Keycloak` calls a `file_get_contents()` when this option key is present in the array of options (see [here](https://github.com/stevenmaguire/oauth2-keycloak/blob/2.0.0/src/Provider/Keycloak.php#L63) and [here](https://github.com/stevenmaguire/oauth2-keycloak/blob/2.0.0/src/Provider/Keycloak.php#L242)), the previously mentioned PHP warning is thrown.

## Proposal

Setting the default value of the optional `encryption_key_path` configuration item for Keycloak clients to `null` avoids this PHP warning.
In this PR, the other optional configuration items are also configured to `null` default values for consistency in this provider.


weaverryan:
Great PR and explanation! Thank you @stephaneseng! I've merged and tagged 0.15.1 :)

PR #72, opened 2018-01-03
Add simple user provider

thewilkybarkid:
This add a user provider that can be used for simple authentication use cases, allowing any successful OAuth to authenticate into your app without any persistence.

(This is similar to [the other bundle](https://github.com/hwi/HWIOAuthBundle/blob/8a97b6b2ba0be0064c9616a3cfc14c2c0d8a589b/Security/Core/User/OAuthUserProvider.php).)

### DIFF
--- a/src/Security/User/OAuthUserProvider.php
+++ b/src/Security/User/OAuthUserProvider.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * OAuth2 Client Bundle
+ * Copyright (c) KnpUniversity <http://knpuniversity.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace KnpU\OAuth2ClientBundle\Security\User;
+
+use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+class OAuthUserProvider implements UserProviderInterface
+{
+    private $roles;
+
+    public function __construct(array $roles = ['ROLE_USER', 'ROLE_OAUTH_USER'])
+    {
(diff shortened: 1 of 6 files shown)


tests/Security/User/OAuthUserProviderTest.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $userProvider = new OAuthUserProvider();
+
+        $this->assertSame($supports, $userProvider->supportsClass($class));
+    }
+
+    public function supportsClassProvider()
+    {
+        yield 'OAuthUser' => [OAuthUser::class, true];
+        yield 'other user' => [SomeUser::class, false];
```
weaverryan: Could you use a `UserInterface` mock instead of creating the stub class below?

src/Security/User/OAuthUserProvider.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+class OAuthUserProvider implements UserProviderInterface
+{
+    private $roles;
+
+    public function __construct(array $roles = ['ROLE_USER', 'ROLE_OAUTH_USER'])
```
weaverryan: I noticed this wasn't actually configurable in the service. And... I'm not sure if it should be anyways :). After all, this is a helper class to make things easy... and not something the user is bound to. With this setup, they could just create their *own* service for this class. I'm just thinking aloud... I *think* like it how it is now

thewilkybarkid:
Had a stab at some documentation, hopefully it's clear enough but please change if not!

weaverryan:
Great job! I made some minor tweaks, but you made my job easy :). I appreciate you taking the time to contribute back - and hopefully you can now make your linked PR a bit simpler - I just tagged 1.16.0.

PR #76, opened 2018-02-10
Fix missing GitlabResourceOwner namespace use

Soullivaneuh:
Note: You should use a tool like PHPStan to prevent this kind of issue. ;-)


weaverryan:
Thanks @Soullivaneuh! Ah, that's a really great idea - would love to automate/eliminate silly errors like this. Would you be willing to open a PR? :) I haven't used the tool before.

Soullivaneuh:
@weaverryan Not time right now but I opened an issue as a remember: #88

PR #79, opened 2018-02-11
Updating Guard docs for Sf4 support

### LINKED ISSUE #77: The guard example documentation should be updated
On Symfony 4.0, if we extends your GuardAuthenticator, we have to implements more method like `start` or `supports`.

For example with this part:

```php
    public function getCredentials(Request $request)
    {
        if ($request->getPathInfo() != '/connect/facebook/check') {
            // don't auth
            return;
        }

        return $this->fetchAccessToken($this->getFacebookClient());
    }
```

The request path check should be done on the `supports` method.

weaverryan:
Fixes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,9 +249,6 @@ use Doctrine\ORM\EntityManager;
 use Symfony\Component\Routing\RouterInterface;
 use KnpU\OAuth2ClientBundle\Client\Provider\FacebookClient;
 use KnpU\OAuth2ClientBundle\Client\ClientRegistry;
-use Symfony\Component\Security\Core\Exception\CustomUserMessageAuthenticationException;
-use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@weaverryan does this not break it for < 4.0 where supports is not checked by symfony?

bocharsky-bw:
It does... I mean, this method just won't be called. But we need to keep things up-to-date with the latest version and keeping different docs for several supported versions would be overkill here I think.

Do you think we need some kind of note here?

alexander-schranz:
@bocharsky-bw maybe something like:

// When using Symfony < 4.0 uncomment the following code:
// if (!$this->supports()) {
//    return;
// }

bocharsky-bw:
Hm, I like it! You want to add it below the `// this method is only called if supports() returns true`, right?

Would you like to create a PR?

alexander-schranz:
@bocharsky-bw its here: https://github.com/knpuniversity/oauth2-client-bundle/pull/97

PR #84, opened 2018-02-17
Fetch GitHub emails from API call if not given on the profile

### LINKED ISSUE #82: Fetching GitHub email from API if not provided on the profile response
Sometime, GitHub does not give the user email on the profile response: https://github.com/thephpleague/oauth2-github/issues/3

IMHO, if the email does not exists, we should call the email API yourself. Otherwise, it will just fail.

Maybe it can be done here: https://github.com/knpuniversity/oauth2-client-bundle/blob/2c1187323e7852279cfc8b982f6ea95715314a7e/src/Client/Provider/GithubClient.php#L31-L34

It was done like that for HWI: https://github.com/hwi/HWIOAuthBundle/pull/758

soullivaneuh:
Closes #82

### DIFF
--- a/src/Client/Provider/GithubClient.php
+++ b/src/Client/Provider/GithubClient.php
@@ -22,7 +22,29 @@ class GithubClient extends OAuth2Client
      */
     public function fetchUserFromToken(AccessToken $accessToken)
     {
-        return parent::fetchUserFromToken($accessToken);
+        $resourceOwner = parent::fetchUserFromToken($accessToken);
+
+        if (null === $resourceOwner->getEmail()) {
+            $response = $this->getOAuth2Provider()->getHttpClient()->request(
+                'GET',
+                'https://api.github.com/user/emails',
+                [
+                    'headers' => [
+                        'Authorization' => "Bearer {$accessToken->getToken()}"
+                    ]
+                ]
+            );
+
+            foreach (json_decode($response->getBody()->getContents(), true) as $email) {
+                if (true === $email['primary']) {
+                    return new GithubResourceOwner(\array_merge($resourceOwner->toArray(), [
+                        'email' => $email['email']
+                    ]));
+                }
+            }
+        }
+
+        return $resourceOwner;
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
     },
     "require-dev": {
         "league/oauth2-facebook": "^1.1|^2.0",
+        "league/oauth2-github": "^0.2.2|^2.0",
         "symfony/security-guard": "^2.8|^3.0",
         "phpunit/phpunit": "^4.8"
     },
--- a/tests/Client/GithubClientTest.php
+++ b/tests/Client/GithubClientTest.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace KnpU\OAuth2ClientBundle\Tests\Client;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Handler\MockHandler;
+use GuzzleHttp\Psr7\Response;
+use KnpU\OAuth2ClientBundle\Client\Provider\GithubClient;
+use League\OAuth2\Client\Provider\GithubResourceOwner;
+use League\OAuth2\Client\Token\AccessToken;
+
+class GithubClientTest extends OAuth2ClientTest
+{
+    public function testFetchUser()
+    {
+        $this->request->request->set('code', 'CODE_ABC');
+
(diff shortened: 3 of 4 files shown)


src/Client/Provider/GithubClient.php
```diff
@@ -22,7 +22,29 @@ class GithubClient extends OAuth2Client
...
+                [
+                    'headers' => [
+                        'Authorization' => "Bearer {$accessToken->getToken()}"
+                    ]
+                ]
+            );
+
+            foreach (json_decode($response->getBody()->getContents(), true) as $email) {
+                if (true === $email['primary']) {
+                    return new GithubResourceOwner(\array_merge($resourceOwner->toArray(), [
```
bocharsky-bw: I think we don't need this `\` in front of `array_merge()`

src/Client/Provider/GithubClient.php
```diff
@@ -22,7 +22,29 @@ class GithubClient extends OAuth2Client
...
+                'https://api.github.com/user/emails',
+                [
+                    'headers' => [
+                        'Authorization' => "Bearer {$accessToken->getToken()}"
+                    ]
+                ]
+            );
+
+            foreach (json_decode($response->getBody()->getContents(), true) as $email) {
+                if (true === $email['primary']) {
```
bocharsky-bw: Are we sure `primary` key always exists on `$email`?  I mean, probably we need to do `isset($email['primary'])` first?

src/Client/Provider/GithubClient.php
weaverryan: It's ok either way. There's a TINY performance boost in some cases with flat functions - so it's kind of hipster right now to add the `\` :p

Soullivaneuh: Tiny performance is always got to take especially on large applications. :-)

BTW, it's just a `\` addition. :wink:

src/Client/Provider/GithubClient.php
Soullivaneuh: According to the [official documentation](https://developer.github.com/v3/users/emails/) and my personal tests, yes this key is always present.

src/Client/Provider/GithubClient.php
```diff
@@ -22,7 +22,29 @@ class GithubClient extends OAuth2Client
      */
     public function fetchUserFromToken(AccessToken $accessToken)
     {
-        return parent::fetchUserFromToken($accessToken);
+        $resourceOwner = parent::fetchUserFromToken($accessToken);
+
+        if (null === $resourceOwner->getEmail()) {
```
weaverryan: We should invert the logic so w can return immediately in the “normal” case - it’s just avoids nesting in the if.

Soullivaneuh:
> This doesn’t seem like the job of this bundle

Not a job for the oauth library, for sure, but can be an extra for this bundle.

In most of the case, the developer need the user e-mail to create an account.

At least, can we propose it as a special option?

> Also, is any specific scope needed to fetch the email?

AFAIK, the `user:email` scope.

bocharsky-bw:
> AFAIK, the `user:email` scope.

IIRC, we do not request this scope out-of-the-box, so it'll always fail by default if the user does not have a public email?

Soullivaneuh:
> so it'll always fail by default if the user does not have a public email?

Exactly! And the authentication will never work.

src/Client/Provider/GithubClient.php
Soullivaneuh: I don't think it will be better because the ressource owner here is returned only if a primary email is found on the foreach.

So I don't have default return if I reverse this condition.

Soullivaneuh:
@weaverryan I'm not sure of what you want. Do you want me to remove all the code modification and explain how to do it on the documentation or leave the code as is but activate only with an option?

weaverryan:
Hmm. Yea, I think this is a docs issue. The proper solution is that if a user needs the email, they should add the `user:email` scope when calling the `->redirect()` method. We do talk about the `redirect()` method in the README, with Facebook as an example. We could add more examples there to make the user really think about it:

```
'public_profile', 'email' // the scopes you want to access
// or, for example, 'user:email' if using GitHub
// check the docs for the exact scopes that you need
```

src/Client/Provider/GithubClient.php
wuestkamp: Scope `user:email` is needed as far as I now

bocharsky-bw:
Can we close this PR as unmerged now? Do we need any tweaks in docs as Ryan suggested?

PR #85, opened 2018-02-17
Remove vilain tabs from services.xml

Soullivaneuh:


bocharsky-bw:
Thanks @Soullivaneuh !

PR #86, opened 2018-02-17
Make http client configurable

### LINKED ISSUE #83: Make collaborators configurable
https://github.com/knpuniversity/oauth2-client-bundle/blob/2c1187323e7852279cfc8b982f6ea95715314a7e/src/DependencyInjection/ProviderFactory.php#L50-L51

I need this to replace the http client (https://github.com/knpuniversity/oauth2-client-bundle/issues/81#issuecomment-366079139), it should be configurable.

P.S.: Todos are bad, use issues instead. :stuck_out_tongue:

Soullivaneuh:
Closes #83

### DIFF
--- a/src/DependencyInjection/KnpUOAuth2ClientExtension.php
+++ b/src/DependencyInjection/KnpUOAuth2ClientExtension.php
@@ -154,6 +154,7 @@ class KnpUOAuth2ClientExtension extends Extension
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('services.xml');
 
+        $httpClient = $config['http_client'];
         $clientConfigurations = $config['clients'];
 
         $clientServiceKeys = [];
@@ -185,6 +186,10 @@ class KnpUOAuth2ClientExtension extends Extension
 
             $configurator = $this->getConfigurator($type);
 
+            $collaborators = [];
+            if ($httpClient) {
+                $collaborators['httpClient'] = new Reference($httpClient);
+            }
             // hey, we should add the provider/client service!
             $clientServiceKey = $this->configureProviderAndClient(
                 $container,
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -248,6 +251,11 @@ private function configureProviderAndClient(ContainerBuilder $container, $provid
             $redirectRoute,
             $redirectParams,
         ]);
+        if (is_string($httpClient)) {
```
weaverryan: Usually we do things like `if (null !== $httpClient)` or just `if ($httpClient)`. It's a bit weird to check for a string specifically.

src/DependencyInjection/ProviderFactory.php
```diff
@@ -39,16 +40,19 @@ public function __construct(UrlGeneratorInterface $generator)
      * @param array $options
      * @param string $redirectUri
      * @param array $redirectParams
+     * @param ClientInterface|null $httpClient
      * @return mixed
      */
-    public function createProvider($class, array $options, $redirectUri, array $redirectParams = [])
+    public function createProvider($class, array $options, $redirectUri, array $redirectParams = [], ClientInterface $httpClient = null)
```
weaverryan: Since there are other potential `collaborators` keys, should we make this new, 4th argument an array called `$collaborators`? And we would, for this PR, set the `ClientInterface` on the `httpClient` key? I'm worried that if we add more collaborator options later, it will get a bit weird to have a 5th & 6th argument, but they're all optional.

src/DependencyInjection/KnpUOAuth2ClientExtension.php
Soullivaneuh: I'll do it, but it's recommended to do strict comparison. I took the habit with PHPStan. :wink:

weaverryan:
Thank you @Soullivaneuh! :)

PR #87, opened 2018-02-17
PHP CS fixer run

Soullivaneuh:
You still use php-cs-fixer v1. Do you want a another PR after this one for upgrade to v2?

Also, it would be great to have cs check enable on CI.

I can provide you an integration with [FlintCI](https://flintci.io) for that. It handles php-cs-fixer v2 and another tools which may help you. :-)


bocharsky-bw:
I agree it would be greate to have CS check on CI as we're talking about Open Source library. Could you send another PR to add PHP-CS-Fixer in TravisCI config file? I think integration with FlintCI is overhead :)

weaverryan:
This is great - thank you @Soullivaneuh!

And +1 for upgrading to version 2 - we should try to keep up to date for sure :).

And +1 for phpcs on CI. I'm not familiar with FlintCI, but I agree with @bocharsky-bw - unless it has some big super powers, I'd prefer to just keep one CI and add php-cs there.

Soullivaneuh:
>  unless it has some big super powers

Well, a tool like FlintCI can comment the PR diff and provide auto-fixing.

Plus, TravisCI job are limited on free plan, moving the CS to FlintCI or other will release some time.

PR #90, opened 2018-02-21
Change underlying library for discord because it's not compliant with oauth2-client v2.x

tristanbes:
Because the previous library is only compatible with the ultra old `thephpleague/oauth2-client` in version` 1.x`.

The documentation no longer reference the library `https://github.com/teamreflex/oauth2-discord`. Instead, they promote https://github.com/wohali/oauth2-discord-new.

See: http://oauth2-client.thephpleague.com/providers/thirdparty/

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ via Composer:
 | [Clever](https://github.com/schoolrunner/oauth2-clever)               | composer require schoolrunner/oauth2-clever         |
 | [DevianArt](https://github.com/SeinopSys/oauth2-deviantart)           | composer require seinopsys/oauth2-deviantart        |
 | [DigitalOcean](https://github.com/chrishemmings/oauth2-digitalocean)  | composer require chrishemmings/oauth2-digitalocean  |
-| [Discord](https://github.com/teamreflex/oauth2-discord)               | composer require team-reflex/oauth2-discord         |
+| [Discord](https://github.com/wohali/oauth2-discord-new)               | composer require team-reflex/oauth2-discord         |
 | [Dribbble](https://github.com/crewlabs/oauth2-dribbble)               | composer require crewlabs/oauth2-dribbble           |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ via Composer:
 | [Clever](https://github.com/schoolrunner/oauth2-clever)               | composer require schoolrunner/oauth2-clever         |
 | [DevianArt](https://github.com/SeinopSys/oauth2-deviantart)           | composer require seinopsys/oauth2-deviantart        |
 | [DigitalOcean](https://github.com/chrishemmings/oauth2-digitalocean)  | composer require chrishemmings/oauth2-digitalocean  |
-| [Discord](https://github.com/teamreflex/oauth2-discord)               | composer require team-reflex/oauth2-discord         |
+| [Discord](https://github.com/wohali/oauth2-discord-new)               | composer require team-reflex/oauth2-discord         |
```
bocharsky-bw: Should be "... require wohali/oauth2-discord-new" in the end, you changed the link URL only

bocharsky-bw:
On one hand, it makes sense to move Discord to the v2, but on the other - we may provide a BC break with this PR. Probably to avoid BC breaks we need to keep both? Or, maybe a bit complicate the logic with some extra checks like if the new class from `wohali/oauth2-discord-new` is exist - use it, otherwise, fallback to the `teamreflex/oauth2-discord`?

tristanbes:
I fixed your comments. good catch. Also, the edits from maintainers are activated if you want to provide any kind of fallbacks (which is overrated imo)

weaverryan:
Thanks @tristanbes!

And yea, I agree - this is a small BC break, but I also really don't want to deal with fallback logic. Hopefully this type of thing will be rare.

tristanbes:
Maybe just add it in UPGRADE or Readme to tell user that they need to require the new lib instead of the previous one. And everything will work the same ;-)

My pleasure.

Issue #39, opened 2016-09-14
Response null given


andrew7030:
Hello, at the outset I want to apologize for my bad English.
I have work to do authentication token JWT and services facebook and google. JWT token works well. But it does not work well this package oauth. I want to do Authenticating with Guard. When a user is logged on Facebook, it's redirected to the login page of Facebook and click the log is returned the error "The controller must return a response (null given). Did you forget to add a return statement somewhere in your controller." The firewall does not work, does not redirect the user even though it was logged on Facebook. Could you look up and see my configurations? I am a bit of a novice in symfony. Best Regards 

security.yml

```yml
    firewalls:
        main:
            pattern: ^/
            anonymous: true
            guard:
                authenticators:
                    - facebook_authenticator
                    - jwt_token_authenticator
                entry_point:  jwt_token_authenticator

    access_control:
        - { path: ^/_wdt|_profiler, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/, roles: IS_AUTHENTICATED_FULLY } 
```

services.yml

```yml
services:
    jwt_token_authenticator:
        class: AppBundle\Security\JwtTokenAuthenticator
        autowire: true

    facebook_authenticator:
        class: AppBundle\Security\FacebookAuthenticator
        autowire: true
...
 ```

connect 
```php
/**
     * @Route("/connect/facebook", name="connect_facebook")
     */
    public function connectAction()
    {
        // will redirect to Facebook!
        return $this->get('oauth2.registry')
            ->getClient('facebook_main') // key used in config.yml
            ->redirect();
    }

    /**
     * @Route("/connect/facebook/check", name="connect_facebook_check")
     */
    public function connectCheckAction(Request $request)
    {
    }
 ```

config.yml

```yml
knpu_oauth2_client:
    clients:
        facebook_main:
            type: facebook
            client_id: %facebook_app_id%
            client_secret: %facebook_app_secret%
            redirect_route: connect_facebook_check
            redirect_params: {}
            graph_api_version: v2.3
```

FacebookAuthenticator

```php
    private $clientRegistry;
    private $em;
    private $router;

    public function __construct(ClientRegistry $clientRegistry, EntityManager $em, RouterInterface $router)
    {
        $this->clientRegistry = $clientRegistry;
        $this->em = $em;
        $this->router = $router;
    }

[...]

PR #96, opened 2018-03-29
Use the router to get the correct path from the connect facebook check route

alexander-schranz:
Prevent problems like: https://github.com/knpuniversity/oauth2-client-bundle/issues/39

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,8 +265,8 @@ class MyFacebookAuthenticator extends SocialAuthenticator
 
     public function supports(Request $request)
     {
-        // continue ONLY if the URL matches the check URL
-        return $request->getPathInfo() == $this->router->getRouteCollection()->get('connect_facebook_check')->getPath();
+        // continue ONLY if the current ROUTE matches the check ROUTE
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -266,7 +266,7 @@ class MyFacebookAuthenticator extends SocialAuthenticator
     public function supports(Request $request)
     {
         // continue ONLY if the URL matches the check URL
-        return $request->getPathInfo() == '/connect/facebook/check';
+        return $request->getPathInfo() == $this->router->getRouteCollection()->get('connect_facebook_check')->getPath();
```
bocharsky-bw: Why don't just `$this->router->generate('connect_facebook_check')`?

alexander-schranz: would add a host when the route has a host defined or?

weaverryan: It shouldn't add a host, but I'm actually not sure if you have the host defined - good point :).

But, there another option anyways, where we look specifically at the *route* name:

```php
return $request->attributes->get('_route') == 'connect_facebook_check';
```

Do you guys like that better?

alexander-schranz: Looks good for me did update the PR

PR #97, opened 2018-03-29
Add hint to call supports in getCredentials method for symfony lower than 3.4

alexander-schranz:
With https://github.com/knpuniversity/oauth2-client-bundle/pull/79 the docs is not longer working for lower than symfony 3.4. So a hint should be added how to get it work with <3.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,8 +273,7 @@ class MyFacebookAuthenticator extends SocialAuthenticator
     {
         // this method is only called if supports() returns true
 
-        // For Symfony < 4.0 the supports method need to be called manually
-        // uncomment the following code or remove it:
+        // For Symfony < 4.0 the supports method need to be called manually:
         // if (!$this->supports()) {
         //    return;
         // }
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks @alexander-schranz! I think it's actually 3.3 and lower that have the issue - 3.4 should act like 4.0. But, I like the idea - we just need to tweak the message.

alexander-schranz:
@weaverryan updated the message to lower than 3.4 and fix the supports function call.

larzuk91:
It looks good to me :+1:

README.md
```diff
@@ -273,6 +273,11 @@ class MyFacebookAuthenticator extends SocialAuthenticator
     {
         // this method is only called if supports() returns true
 
+        // For Symfony lower than 3.4 the supports method need to be called manually here:
+        // if (!$this->supports($request)) {
+        //    return;
```
bocharsky-bw: Let's explicitly return `null` here since this method is supposed to return something, but just `return;` is supposed to return `void`

weaverryan:
Thank you again!

PR #105, opened 2018-06-20
Added a hint about the relation to hwi/oauth-bundle

Toflar:
Feel free to edit and fix my English 😄 

Also, I didn't want to write something like `Ideally this bundle is eventually going to be replaced by hwi/oauth-bundle` because I don't know your plans for it. But you might want to add some words about the future of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,12 @@ This bundle integrates with [league/oauth2-client](http://oauth2-client.thephple
 
 * PHP 5.5.9 or higher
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Boom! I made some tweaks, but you made it easy for me. Thanks for this great idea :)

Toflar:
Wow, the tweaks make it so easier to read 🤣 Cheers!

PR #106, opened 2018-06-26
Remove state from the session

thewilkybarkid:
This removes the state from the session once it's read, as it's no longer needed.


weaverryan:
Thank you @thewilkybarkid!

PR #114, opened 2018-08-22
Add configurable option to use oidc mode for the Google Provider

hansnilsson:
The PHP league Google client has an option named useOidcMode which this PR lets you set in this bundle configuration.

See https://github.com/thephpleague/oauth2-google#resource-owner-attributes


weaverryan:
Wow, thank you! Great PR and you even linked to the source so I could understand. Cheers!

Issue #104, opened 2018-06-07
"Auth0 account is not specified"


picks44:
Hello,

I'm trying to connect through Auth0, but I get the error  `Auth0 account is not specified` from the `riskio/oauth2-auth0` bundle when I try to reach `/connect/auth0` route.

What is wrong ?

My setup is the following (Symfony 3.4)
```
knpu_oauth2_client:
    clients:
        auth0:
            type: auth0
            client_id: '%auth0_client_id%' (in parameters.yml)
            client_secret: '%auth0_client_secret%' (in parameters.yml)
            redirect_route: connect_auth0_check
            redirect_params: {}
```

```
    /**
     * @Route("/connect/auth0")
     *
     * @return \Symfony\Component\HttpFoundation\RedirectResponse
     * @throws \InvalidArgumentException
     */
    public function connectAction()
    {
        return $this->get('oauth2.registry')
            ->getClient('auth0')
            ->redirect();
    }

    /**
     * @Route("/connect/auth0/check", name="connect_auth0_check")
     *
     * @param Request $request
     * @throws \InvalidArgumentException
     */
    public function connectCheckAction(Request $request)
    {
        $client = $this->get('oauth2.registry')
            ->getClient('auth0');

        try {
            $user = $client->fetchUser();
            var_dump($user->getName());
        } catch (IdentityProviderException $e) {
            var_dump($e->getMessage());die;
        }
    }
```

picks44:
@weaverryan can you help here ? :smiley:

weaverryan:
Hey!

Hmm, I'm not sure. Their code makes it seem like there's an undocumented "account" option that's used in the URL - https://github.com/RiskioFr/oauth2-auth0/blob/master/src/Auth0.php#L18-L22

If you try to follow their README directly - actually creating the client by hand (https://github.com/RiskioFr/oauth2-auth0/tree/master/src) does it work? Or do you still get this error? If you DO get this error still, try passing an `account` option to the client (in addition to clientId, clientSecret, etc). If that works, then we'll know there is an undocumented option that we need to add to this bundle (and probably we need to update their docs also).

8eph:
@weaverryan @picks44 

This should work swimmingly. The only thing I'm uncertain about is whether the config setting should be mandatory or not.

I'm assuming this worked at one point even without 'account', however this is currently not the case.

PR #117, opened 2018-08-23
Missing config key for Auth0

8eph:
Add mandatory config key 'account' for Auth0
Update README

Requested in #104


weaverryan:
Thank you @8eph  - really nice job! I agree with making it required - I don't see how it could ever be optional :).

PR #118, opened 2018-08-24
Using simpler getMessage(), which is identical

### LINKED ISSUE #115: Call to a member function getMessage() on null
Hello,

Running this `knpuniversity/oauth2-client-bundle` on production gives sometimes error like the following: 

bug

```
Error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: 
Call to a member function getMessage() on null in
 <zap>/vendor/knpuniversity/oauth2-client-bundle/src/Security/Exception/IdentityProviderAuthenticationException.php:34
```

I am on version `1.18.0`

Am I doing something wrong ? Or could it be possible that `IdentityProviderAuthenticationException` sometimes has no `getPrevious()` object returned ? (in which case we need the code to be more defensive)

weaverryan:
Fixes #115

### DIFF
--- a/src/Security/Exception/IdentityProviderAuthenticationException.php
+++ b/src/Security/Exception/IdentityProviderAuthenticationException.php
@@ -31,7 +31,7 @@ class IdentityProviderAuthenticationException extends AuthenticationException
     public function getMessageData()
     {
         return [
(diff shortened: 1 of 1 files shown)


tristanbes:
Thank you 👍

Issue #130, opened 2018-09-26
Support more options for azure


aeimer:
Hi there,

I wanted to set the `tenant` for my `azure` Adapter. The adapter supports it, but I couldn't configure it over the yaml.

I had to use the `generic` adapter to pass some extra options to the provider (`provider_options`).

Would it be possible to make these things configurable via the azure provider? It would make the usage more easy and may save future users a lot of time...

Example adaption for the Configurator would be:

```php
<?php

/*
 * OAuth2 Client Bundle
 * Copyright (c) KnpUniversity <http://knpuniversity.com/>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;

use Symfony\Component\Config\Definition\Builder\NodeBuilder;

class AzureProviderConfigurator implements ProviderConfiguratorInterface
{
    public function buildConfiguration(NodeBuilder $node)
    {
        $node
            ->scalarNode('url_login')
                ->info('TODO')
            ->end()
            ->scalarNode('tenant')
                ->info('TODO')
            ->end()
            ->scalarNode('api_version')
                ->info('TODO')
            ->end()
            ->scalarNode('url_api')
                ->info('TODO')
            ->end()
        // There are several more options, I'd be fine with having these basic ones
        // See https://github.com/TheNetworg/oauth2-azure/blob/master/src/Provider/Azure.php
    }

    public function getProviderClass(array $config)
    {
        return 'TheNetworg\OAuth2\Client\Provider\Azure';
    }

    public function getProviderOptions(array $config)
    {
        return [
            'clientId' => $config['client_id'],
            'clientSecret' => $config['client_secret'],
            'urlLogin' => $config['url_login'],
            'urlAPI' => $config['url_api'],
            'tenant' => $config['tenant'],
            'API_VERSION' => $config['api_version'],
        ];
    }

    public function getPackagistName()
    {
        return 'thenetworg/oauth2-azure';
    }

    public function getLibraryHomepage()
    {
        return 'https://github.com/thenetworg/oauth2-azure';
    }

    public function getProviderDisplayName()
    {
        return 'Azure';
    }

    public function getClientClass(array $config)
    {
        return 'KnpU\OAuth2ClientBundle\Client\Provider\AzureClient';
    }
}

```

aeimer:
@bocharsky-bw done ;)

[...]

PR #131, opened 2018-10-04
Adds whole config for azure

aeimer:
Adds all configurable vars to tree builder.

https://github.com/knpuniversity/oauth2-client-bundle/issues/130


aeimer:
I found out that `$urlAPI` and `$API_VERSION` doesnt't seem to be used anywhere. So we could remove them from the config...

@hajekj do you have any comment on this?

weaverryan:
Thanks guys!

PR #135, opened 2018-10-12
Feature/#133 drupal provider base url not configurable

mnugter:
Adds base_url as a configurable property and update the README configuration example to reflect the option.


larzuk91:
Code looks good :+1:

PR #137, opened 2018-10-31
Added Oauth2ClientInterface for service decorating

### LINKED ISSUE #134: Add the possibility to retrieve services via DI
Hi there,

it would be great to have a way to inject services to the classes. 

Is it possible to make a `$container` variable available for most of the Interfaces etc?

The reason would be to use eg logging and caching in the classes. Currently there is no way I know to load those things.

Best regards

sadikoff:
Example:
```php

namespace App\Client;

class CacheableClient implements OAuth2ClientInterface
{
    private $client;
    private $cache;

    public function __construct(OAuth2ClientInterface $client, Cache $cache)
    {
        // ...
    }

    // override all public functions, call the method on the internal $this->client object
    // but add caching wherever you need it
}
```

```yaml
services:
    app.azure_client:
        class: App\Client\CacheableClient
        decorates: knpu.oauth2.client.azure
```
closes #134

### DIFF
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -18,7 +18,7 @@ use League\OAuth2\Client\Token\AccessToken;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
 
-class OAuth2Client
+class OAuth2Client implements OAuth2ClientInterface
 {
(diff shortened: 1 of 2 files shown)


weaverryan:
@sadikoff This looks perfect! Could you also add a small section in the README file, maybe call it "Extending/Decorating Client Classes"?

sadikoff:
@weaverryan added, but it needs your review.

PR #141, opened 2018-11-14
Update Azure DI Config With Default Values

brianfreytag:
The update to add all of the Azure-specific configuration variables into
the tree builder did not set any defaults, so you have to configure each
and every option in the yaml file.

This updates #131 to include the proper default values as defined in the
TheNetworg/oauth2-azure Provider class. This way configuration can be as
minimal as possible, allowing you to simply overwrite default values
instead of defining each and every one.


src/DependencyInjection/Providers/AzureProviderConfigurator.php
```diff
@@ -35,21 +39,26 @@ public function buildConfiguration(NodeBuilder $node)
...
+                ->defaultValue('common')
             ->end()
             ->scalarNode('url_api')
                 ->example("url_api: 'https://graph.windows.net/'")
                 ->info('Domain to build request URL')
+                ->defaultValue('https://graph.windows.net/')
             ->end()
             ->scalarNode('resource')
                 ->info('Oauth resource field')
+                ->defaultValue('')
```
weaverryan: Is this required? I mean, is empty quotes different than allowing this just to be null?

src/DependencyInjection/Providers/AzureProviderConfigurator.php
```diff
@@ -35,21 +39,26 @@ public function buildConfiguration(NodeBuilder $node)
...
             ->end()
             ->scalarNode('api_version')
                 ->example("api_version: '1.6'")
                 ->info('The API version to run against')
+                ->defaultValue('1.6')
```
weaverryan: I think we should avoid this. The problem is that setting a default API version means we may need to "bump" it up in the future. Also, making the user choose is a good idea - then they will go find the correct version of the docs for that version.

src/DependencyInjection/Providers/AzureProviderConfigurator.php
brianfreytag: I'm just following what the Provider has set as their defaults:

```
class Azure extends AbstractProvider
{
    use BearerAuthorizationTrait;
    public $urlLogin = 'https://login.microsoftonline.com/';
    public $pathAuthorize = '/oauth2/authorize';
    public $pathToken = '/oauth2/token';
    public $scope = [];
    public $scopeSeparator = ' ';
    public $tenant = 'common';
    public $urlAPI = 'https://graph.windows.net/';
    public $resource = '';
    public $API_VERSION = '1.6';
    public $authWithResource = true;
...
}
```

I gave TheNetworg devs the benefit of the doubt. I figured they know what they're doing, so there may be a reason it needs to be an empty string as opposed to a null value.

src/DependencyInjection/Providers/AzureProviderConfigurator.php
brianfreytag: I concur, with reference to "we may need to "bump" it up". I wouldn't want to be responsible for watching the provider to see if they change their default API version.

That being said, we could run into the same issue with any of these configuration options. I mean, the odds of Microsoft making significant changes to its default paths, login url, graph url, etc are very slim. I doubt they want to break BC that bad, but it could happen and provider would end up needing updated leading us to have to update it here. 

API version I can see being a problem; However, the initial commit was November 16, 2015 and it was 1.6 then as well, so idk. I might push back a bit on the `resource` empty string one above, but this one -- just tell me what to do and I'll do it.

src/DependencyInjection/Providers/AzureProviderConfigurator.php
brianfreytag: I can confirm that `null` value for resource works (I just went into my project and modified the config and the defaults in the class and authentication still works. I will update this to default to `null`.

brianfreytag: I updated the  `'resource'` node  with `->defaultNull()`

src/DependencyInjection/Providers/AzureProviderConfigurator.php
brianfreytag: I updated the `'api_version'` node to remove the default value altogether

brianfreytag:
@weaverryan what are your thoughts on this?

weaverryan:
Thanks for the ping! Merged and v1.25.0 tagged.

brianfreytag:
Thank you, kindly, sir.

PR #149, opened 2018-12-08
Fix README link to "Guard authenticator" screencast

techdaddies-kevin:
Fixes the knpuniversity.com link to point to symfonycasts.com screencast on symfony-security. Also adds the URL protocol since current link results in a 404 on Github


weaverryan:
Wow, yes - thank you @techdaddies-kevin!

PR #151, opened 2018-12-14
added ability to configure default http_client

sadikoff:
Added to config new options
```yaml
http_client_options:
    timeout: '30' # seconds 
    proxy: 'proxy_url' # proxy url string
    verify: true|false # verify or not ssl certificate
```

Fixed some bugs
* symfony 4.2 deprecation
* wrong cache & log dirs in tests
* missed `%s` in `InvalidConfigurationException` message

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -17,12 +17,20 @@ class Configuration implements ConfigurationInterface
 {
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('knpu_oauth2_client');
+        $treeBuilder = new TreeBuilder('knpu_oauth2_client');
+        $rootNode = method_exists($treeBuilder, 'getRootNode') ? $treeBuilder->getRootNode() : $treeBuilder->root('knpu_oauth2_client');
 
         $rootNode
             ->children()
             ->scalarNode('http_client')->defaultNull()->info('Service id of HTTP client to use (must implement GuzzleHttp\ClientInterface)')->end()
+            ->arrayNode('http_client_options')
+                ->info('')
+                ->children()
+                    ->integerNode('timeout')->min(0)->end()
+                    ->scalarNode('proxy')->end()
+                    ->booleanNode('verify')->info('Use only with proxy option set')->end()
+                ->end()
+            ->end()
             ->arrayNode('clients')
                 ->normalizeKeys(false)
                 ->useAttributeAsKey('variable')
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -155,14 +155,15 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.xml');
 
         $httpClient = $config['http_client'];
+        $httpClientOptions = array_key_exists('http_client_options', $config) ? $config['http_client_options'] : [];
```
weaverryan: The array key should always exist - because it's in the `Configuration` class

tests/app/TestKernel.php
```diff
@@ -61,4 +61,22 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
+    }
+
+    public function getLogDir()
+    {
+        if (method_exists($this, 'getProjectDir')) {
+            return $this->getProjectDir() . '/tests/app/cache/' . $this->getEnvironment();
+        }
+
+        return parent::getLogDir();
+    }
```
weaverryan: I assume this fixes some deprecation? Can you explain a bit more?

src/DependencyInjection/Configuration.php
```diff
@@ -17,12 +17,20 @@ class Configuration implements ConfigurationInterface
...
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('knpu_oauth2_client');
+        $treeBuilder = new TreeBuilder('knpu_oauth2_client');
+        $rootNode = method_exists($treeBuilder, 'getRootNode') ? $treeBuilder->getRootNode() : $treeBuilder->root('knpu_oauth2_client');
 
         $rootNode
             ->children()
             ->scalarNode('http_client')->defaultNull()->info('Service id of HTTP client to use (must implement GuzzleHttp\ClientInterface)')->end()
+            ->arrayNode('http_client_options')
```
weaverryan: We need some validation to prevent you from setting both `http_client` and `http_client_options`.

src/DependencyInjection/Configuration.php
```diff
@@ -17,12 +17,20 @@ class Configuration implements ConfigurationInterface
...
+                ->info('')
```
weaverryan: Is this needed?

tests/app/TestKernel.php
sadikoff: this is not deprecation fix, but just fix. With latest vendors after running test suite I got cache and logs dir not in their places. After some investigation I found that there were changes Symfony HttpKernel and we need tune this place

src/DependencyInjection/Configuration.php
weaverryan: I think we still need this one part. I think you may need to add some validation to the parent node - here's an example I think will work https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L630-L635

PR #153, opened 2018-12-23
added fields to odnoklassniki provider

OtezVikentiy:
According to this documentation https://apiok.ru/dev/methods/ There are more fields needed

### DIFF
--- a/src/DependencyInjection/KnpUOAuth2ClientExtension.php
+++ b/src/DependencyInjection/KnpUOAuth2ClientExtension.php
@@ -325,6 +325,8 @@ class KnpUOAuth2ClientExtension extends Extension
         $optionsNode
             ->scalarNode('client_id')->isRequired()->end()
             ->scalarNode('client_secret')->isRequired()->end()
+            ->scalarNode('client_public')->defaultNull()->end()
+            ->scalarNode('access_token')->defaultNull()->end()
             ->scalarNode('redirect_route')->isRequired()->end()
             ->arrayNode('redirect_params')
                 ->prototype('scalar')->end()
--- a/src/DependencyInjection/KnpUOAuth2ClientExtension.php
+++ b/src/DependencyInjection/KnpUOAuth2ClientExtension.php
@@ -325,8 +325,6 @@ class KnpUOAuth2ClientExtension extends Extension
         $optionsNode
             ->scalarNode('client_id')->isRequired()->end()
             ->scalarNode('client_secret')->isRequired()->end()
-            ->scalarNode('client_public')->defaultNull()->end()
-            ->scalarNode('access_token')->defaultNull()->end()
             ->scalarNode('redirect_route')->isRequired()->end()
             ->arrayNode('redirect_params')
                 ->prototype('scalar')->end()
--- a/src/DependencyInjection/Providers/OdnoklassnikiProviderConfigurator.php
+++ b/src/DependencyInjection/Providers/OdnoklassnikiProviderConfigurator.php
@@ -29,6 +29,8 @@ class OdnoklassnikiProviderConfigurator implements ProviderConfiguratorInterface
         return [
             'clientId' => $config['client_id'],
             'clientSecret' => $config['client_secret'],
+            'clientPublic' => $config['client_public'],
+            'accessToken' => $config['access_token'],
         ];
     }
 


src/DependencyInjection/Providers/OdnoklassnikiProviderConfigurator.php
```diff
@@ -29,6 +29,8 @@ public function getProviderOptions(array $config)
         return [
             'clientId' => $config['client_id'],
             'clientSecret' => $config['client_secret'],
+            'clientPublic' => $config['client_public'],
+            'accessToken' => $config['access_token'],
```
weaverryan: Looking at the library - https://github.com/rakeev/oauth2-odnoklassniki - I think only `clientPublic` is needed - `accessToken` is something you get back after the user is redirected back.

src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -325,6 +325,8 @@ private function buildConfigurationForType(NodeDefinition $node, $type)
         $optionsNode
             ->scalarNode('client_id')->isRequired()->end()
             ->scalarNode('client_secret')->isRequired()->end()
+            ->scalarNode('client_public')->defaultNull()->end()
+            ->scalarNode('access_token')->defaultNull()->end()
```
weaverryan: Instead of putting this here, add it to `OdnoklassnikiProviderConfigurator:: buildConfiguration()` - that's where you can add config options that are custom to this provider.

src/DependencyInjection/Providers/OdnoklassnikiProviderConfigurator.php
OtezVikentiy: unfortunately accessToken is also needed acording to documentation.i think that https://github.com/rakeev/oauth2-odnoklassniki has not been updated for a long time, so it does not work properly. I've tested this solution and opened pull request also for it here https://github.com/rakeev/oauth2-odnoklassniki/pull/4/files

But there are problems with tests

src/DependencyInjection/KnpUOAuth2ClientExtension.php
OtezVikentiy: removed these changes

src/DependencyInjection/Providers/OdnoklassnikiProviderConfigurator.php
OtezVikentiy: access token is not a generated value. it is static and provided by ok.ru when you register for using API

weaverryan: Hmm, I think you might be getting something confused related to the accessToken. The purpose of the OAuth2 flow is to *obtain* the accessToken for a specific user. And *then* you use that access token to make API requests to, for example, get user information. 

> access token is not a generated value. it is static and provided by ok.ru when you register for using API

 It *is* possible that you're able to generate an access token for your *own* user (GitHub allows this). If you simply want to make API requests using a hardcoded access token, then you don't need the OAuth "flow": you can just make API requests directly and add the `?access_token=` - I wouldn't even use this library or that other library if you have that situation.

OtezVikentiy: I'm sure that access_token is neended:
1) API of ok.ru fucksUp without this param
2) it is strictly required even in test queries on the site ok.ru
3) it's not a generated value and it is not returned by the API to each user
4) i've tested this solution and it works.... o_O

weaverryan: > I'm sure that access_token is neended:

It's definitely needed to access the API - completely agree. But I've never heard of an OAuth server that required the access_token during the *authorization* step. The whole point of that step is to *get* the `access_token`. Can you point me to their docs to show your point? Or maybe post some functional code that shows you passing the `access_token` in the URL during the step when you're redirecting the user to ok.ru?

weaverryan: Yea, that sounds crazy - it makes no sense at all :). But, whatever, if it works that way, that's fine. But, even if I merge this, you wouldn't be able to use it until your PR is merged: https://github.com/rakeev/oauth2-odnoklassniki/pull/4

I might recommend using the `generic` provider instead so you can wire it up however you want - it doesn't look like your PR is getting much attention.

OtezVikentiy: Yeah, of course i understand this )) I'm on progress with the second PullRequest, i've got a problem running tests locally at the moment (((

WarGot-by:
Any news ?

OtezVikentiy:
> Any news ?
> 
> @OtezVikentiy @weaverryan

Unfortunately i haven't got enough time for this right now, but i'm planning to solve this problem in the nearest future )))

WarGot-by:
Got it. Thanks. I'll try to find the time to fix it myself.

> > Any news ?
> > @OtezVikentiy @weaverryan
> 
> Unfortunately i haven't got enough time for this right now, but i'm planning to solve this problem in the nearest future )))

Got it. Thanks. I'll try to find the time to fix it myself.

weaverryan:
Closing this for now. Please re-open if/when it's ready (I think we were waiting on a PR on another library).

Cheers!

PR #167, opened 2019-03-20
Adding "api_version" and "fields" options for Linkedin's provider

Sicaa:
Hi,

I ran into some issues implementing Linkedin OAuth, basically because:

1. Linkedin is switching from v1 to v2 of their APIs, and there is no way to specify version to use from this bundle. Since all new Linkedin API project cannot use v1, OAuth is unusable with this bundle. PHPLeague oauth2-linkedin provider provide a withResourceOwnerVersion() method since 4.1.0 to set version to use, and I'm planning to submit a PR with this feature in this repo, but said that, I must take one thing at a time :)

2. In conjunction of their version upgrade, Linkedin enforced their policy, and some of the default fields retrieved from PHPLeague provider are now forbidden without requesting developer authorization, so after authentication, it's all 403 responses (=> "Not enough permissions to access: GET /me"). Furthermore, fields that are still available from v1 to v2 changed names ("first-name" becomes "firstName" for example). Fortunately, PHPLeague Linkedin provider allows to overload retrieved fields, so here is my PR!

If you have any suggestions, do not hesitate since I'm new to open-source contributions ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -950,6 +950,8 @@ knpu_oauth2_client:
             # a route name you'll create
             redirect_route: connect_linkedin_check
             redirect_params: {}
+            # Optional value to specify Linkedin's API version to use. As the time of writing, v1 is still used by default by league/oauth2-linkedin.
+            # api_version: null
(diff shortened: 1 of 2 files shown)


Sicaa:
Finally I was able to make the changes for API version today and I thought it would be easier to review in the same PR, so here it is!

weaverryan:
This looks great! Thanks for the detailed outline and great PR!

PR #170, opened 2019-03-27
Fix provider order (minor)

umulmrum:
Sorry, couldn't leave the list almost-sorted :-)


weaverryan:
Yea, the fix here would be to update that script to make them alphabetical! I'm not sure why it's currently *almost* ordered :)

umulmrum:
Ah I missed that - OK, changed.

weaverryan:
Thank you @umulmrum!

PR #171, opened 2019-03-28
Tweak README with bin/update_readme command

bocharsky-bw:
After running `bin/update_readme` the repo is in a dirty state, this PR fixes it.

P.S. I opened #173 to prevent such cases.


weaverryan:
Thank you :)

PR #172, opened 2019-03-28
 Use NULL instead of empty string for defaultNull() values in README

bocharsky-bw:
What about to use `NULL` instead of empty strings? It's clearer I think because in config definition we use `defaultNull()` calls for it.


weaverryan:
GREAT improvement - good attention to detail 💯

PR #173, opened 2019-03-28
Fail CI if repo is in dirty state after bin/update_readme

bocharsky-bw:
This will help us to catch inconsistence in README on CI level. No more missing call of `bin/update_readme`!

The build should be green after #171 is merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,6 @@ before_install:
 install: composer update --prefer-dist $COMPOSER_FLAGS
 
 # Fail CI if repo is in dirty state after bin/update_readme
-before_script: ./bin/update_readme && git diff --exit-code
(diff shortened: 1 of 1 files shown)


weaverryan:
It’s merged! Should we rebase to verify the green build?

bocharsky-bw:
Yeah, I want to do this! Just waiting for tests to actually fail. Unfortunately, Travis Status says that they have slow booting, that's why my builds are pending (yellow) and I don't see it's actually failed right now :/

larzuk91:
They indeed failed

bocharsky-bw:
I think this one is ready now! I merged master and now tests should pass because we fixed dirty state in master

weaverryan:
Thanks! Awesome idea!

PR #193, opened 2019-09-12
Errata in classname for authenticator

rodrigobb:
Classname for Facebook Authenticator in security.yaml seems to have an errata


bocharsky-bw:
Thank you for this fix @rodrigobb !

PR #204, opened 2019-10-05
Phpstan fixes

alister:
A group of individual commits each fixing one distinct issue.

They are:
- Ignoring one class of error that is harder to check (DI config)
- Checking that a service exists before using it
- Asserting particular types are being dealt with
- Returning the proper type - strings, not bool
- Docblock additions & tweaks to suggest more generally, what is actually being returned

### DIFF
--- a/src/Client/ClientRegistry.php
+++ b/src/Client/ClientRegistry.php
@@ -41,15 +41,23 @@ class ClientRegistry
      */
     public function getClient($key)
     {
-        if (!isset($this->serviceMap[$key])) {
-            throw new \InvalidArgumentException(sprintf(
-                'There is no OAuth2 client called "%s". Available are: %s',
-                $key,
-                implode(', ', array_keys($this->serviceMap))
-            ));
+        if (isset($this->serviceMap[$key])) {
+            $client = $this->container->get($this->serviceMap[$key]);
+            if (!$client instanceof OAuth2Client) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Somehow the "%s" client is not an instance of OAuth2Client.',
+                    $key
+                ));
+            }
+
+            return $client;
         }
 
-        return $this->container->get($this->serviceMap[$key]);
+        throw new \InvalidArgumentException(sprintf(
+            'There is no OAuth2 client called "%s". Available are: %s',
+            $key,
+            implode(', ', array_keys($this->serviceMap))
+        ));
     }
 
     /**
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -83,7 +83,7 @@ class OAuth2Client implements OAuth2ClientInterface
     /**
      * Call this after the user is redirected back to get the access token.
      *
-     * @return \League\OAuth2\Client\Token\AccessToken
+     * @return AccessToken|\League\OAuth2\Client\Token\AccessTokenInterface
      *
      * @throws InvalidStateException
      * @throws MissingAuthorizationCodeException
@@ -132,6 +132,7 @@ class OAuth2Client implements OAuth2ClientInterface
      */
     public function fetchUser()
     {
+        /** @var AccessToken $token */
         $token = $this->getAccessToken();
 
         return $this->fetchUserFromToken($token);
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -167,12 +167,10 @@ class OAuth2Client implements OAuth2ClientInterface
      */
     private function getSession()
     {
-        $session = $this->getCurrentRequest()->getSession();
-
-        if (!$session instanceof SessionInterface) {
+        if (!$this->getCurrentRequest()->hasSession()) {
             throw new \LogicException('In order to use "state", you must have a session. Set the OAuth2Client to stateless to avoid state');
         }
 
-        return $session;
+        return $this->getCurrentRequest()->getSession();
     }
 }
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -17,6 +17,7 @@ use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Token\AccessToken;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 
 class OAuth2Client implements OAuth2ClientInterface
 {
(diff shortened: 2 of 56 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -18,9 +18,11 @@ class Configuration implements ConfigurationInterface
...
     {
         $treeBuilder = new TreeBuilder('knpu_oauth2_client');
-        $rootNode = method_exists($treeBuilder, 'getRootNode')
-            ? $treeBuilder->getRootNode()
-            : $treeBuilder->root('knpu_oauth2_client');
+        if (method_exists($treeBuilder, 'root')) {
+            $rootNode = $treeBuilder->root('knpu_oauth2_client'); // BC for Symfony < 4.2
+        } else {
+            $rootNode = $treeBuilder->getRootNode();
+        }
```
weaverryan: What caused this change? I think it's not correct. The `root()` method is deprecated, but (for example) it exists in Symfony 4.3. So, this would cause a user in 4.3/4.4 to use the `root()` method and see a deprecation warning. We need to check for the presence of `getRootNode()`, which is the newer method.

src/DependencyInjection/KnpUOAuth2ClientExtension.php
```diff
@@ -184,9 +184,11 @@ public function load(array $configs, ContainerBuilder $container)
...
             // process the configuration
             $tree = new TreeBuilder('knpu_oauth2_client/clients/'.$key);
-            $node = method_exists($tree, 'getRootNode')
-                ? $tree->getRootNode()
-                : $tree->root('knpu_oauth2_client/clients/'.$key);
+            if (method_exists($tree, 'root')) {
+                $node = $tree->root('knpu_oauth2_client/clients/'.$key); // BC for Symfony < 4.2
+            } else {
+                $node = $tree->getRootNode();
+            }
```
weaverryan: Same here

src/Client/ClientRegistry.php
```diff
@@ -41,15 +41,18 @@ public function __construct(ContainerInterface $container, array $serviceMap)
...
-            throw new \InvalidArgumentException(sprintf(
-                'There is no OAuth2 client called "%s". Available are: %s',
-                $key,
-                implode(', ', array_keys($this->serviceMap))
-            ));
+        if (isset($this->serviceMap[$key])) {
+            $client = $this->container->get($this->serviceMap[$key]);
+            if ($client instanceof OAuth2Client) {
+                return $client;
+            }
```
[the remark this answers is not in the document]
alister: I did wonder about that. The exception message would then be `'The service called "%s" is not an OAuth2 client. Available are: %s'` with the same $key and imploded `array_keys($this->serviceMap)`.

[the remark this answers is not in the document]
weaverryan: It's such an insane error... it basically would only happen if there was a bug in the bundle itself. So, we can simplify to:

> Somehow the "%s" client is not an instance of OAuth2Client.

src/Client/OAuth2Client.php
```diff
@@ -131,6 +132,7 @@ public function fetchUserFromToken(AccessToken $accessToken)
      */
     public function fetchUser()
     {
+        /** @var AccessToken $token */
```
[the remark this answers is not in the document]
alister: Even with the |AccessToken, the AbstractProvider::getAccessToken that OAuth2Client::getAccessToken calls still only returns an AccessTokenInterface. However, OAuth2Client::fetchUserFromToken as well as the AbstractProvider::getResourceOwner that it calls, take the hard classes. 

In 7.4 the types can be wider/narrower where it makes sense, but in the meantime, we have to suggest to PHPStan that it's it's also the hard class that fetchUserFromToken expects.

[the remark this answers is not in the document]
weaverryan: Yea, that's a mess. I guess checking if it's not an instance of `AccessToken` and throwing an exception is more honest?

src/DependencyInjection/Configuration.php
alister: I was trying to avoid adding more PHPStan `ignoreErrors`, but since that call must be there for <4.2, I think that it is inevitable for now.

src/Client/OAuth2Client.php
```diff
@@ -82,7 +83,7 @@ public function redirect(array $scopes = [], array $options = [])
     /**
      * Call this after the user is redirected back to get the access token.
      *
-     * @return \League\OAuth2\Client\Token\AccessToken
+     * @return \League\OAuth2\Client\Token\AccessTokenInterface
```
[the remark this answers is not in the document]
alister: It won't help phpstan (it may read through it and see it's returning a AccessTokenInterface), but it may make what is happening a tad clearer for the rest of us.

src/Client/OAuth2Client.php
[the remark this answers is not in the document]
alister: It could only be one of those two, depending on the version and so if the interface was available (and described only in the docblock return types). The tree of code it runs from that point should only be methods described by the Interface, which is perfectly handled by the AccessToken class anyway. 

So, the comment does for PHPStan, what PHP 7.4 would be able to do for itself - widening the acceptable type.

src/Security/Helper/FinishRegistrationBehavior.php
```diff
@@ -31,10 +32,13 @@ trait FinishRegistrationBehavior
...
-            'guard.finish_registration.user_information',
-            $e->getUserInformation()
-        );
+        $session = $request->getSession();
+        if ($session instanceof SessionInterface) {
+            $session->set(
+                'guard.finish_registration.user_information',
+                $e->getUserInformation()
+            );
+        }
```
bocharsky-bw: What if not? Probably better to throw an exception else? And the same in a few other places. Wdyt?

alister: indeed - OAuth2Client::getSession() will throw a \LogicException if there is no session - and it's 'stateless''. Should a session be required everywhere - and how much stateless-ness is needed?

weaverryan: Hmm, in this case, because the user would be calling this method directly (and the method is called `saveUserInfoToSession()`), if we're unable to actually save to the session because there is none, I think an exception is a good idea

alister: Taking a look in the `Request` class, Symfony 4.1 deprecated calling `getSession()` if there one didn't exist - it's been triggering a notice since then, and 5.0 will throw a `BadMethodCallException`.  Till 4.4 or so though, it was  however, still marked as potentially returning null.

I'll make a friendlier error message instead for all the checks with a `if (!$request->hasSession()) { throw .... }`.  The`hasSession()` method was in for our minimum-supported version of 3.4.

weaverryan:
Thank you @alister! :)

PR #206, opened 2019-10-15
added the custom_domain parameter to the auth0 client library config

smatyas:
This PR makes possible to use the the auth0 client library with a custom domain setting.
It depends on this PR in the client library: https://github.com/RiskioFr/oauth2-auth0/pull/16

The `account` parameter is not required anymore, as the `custom_domain` parameter takes priority if it's set. It is implemented in a backward compatible way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,8 +470,12 @@ knpu_oauth2_client:
             # a route name you'll create
             redirect_route: connect_auth0_check
             redirect_params: {}
+            # Your custom/definite Auth0 domain, e.g. "login.mycompany.com". Set this if you use Auth0's Custom Domain feature. The "account" and "region" parameters will be ignored in this case.
+            custom_domain: null
             # Your Auth0 domain/account, e.g. "mycompany" if your domain is "mycompany.auth0.com"
             account: null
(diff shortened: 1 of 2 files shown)


smatyas:
Exactly, we have to wait until that gets merged.

In the meantime, what should I do about the failing check? Just apply that diff I can see in travis?

bocharsky-bw:
Ah, yes, exactly! You should apply that diff manually, or even better just run the `./bin/update_readme` that should do it for you - we need to update the README according to your changes to make Travis happy

smatyas:
Alright, I've updated the readme using that command.

So let's just wait until https://github.com/RiskioFr/oauth2-auth0/pull/16 is merged, then I'll ping this PR again.

Thanks for your fast replies! :)

smatyas:
OK, the https://github.com/RiskioFr/oauth2-auth0/pull/16 got merged without any major changes (i.e. parameter name). It is already released in v2.2.0, so we can move forward with this PR as well.

Should I do anything more here? Rebase/squash/etc.?

bocharsky-bw:
Thank you for this ping! I think this one is ready and doesn't need any changes 👍

weaverryan:
Thanks @smatyas! And released!

Issue #159, opened 2019-01-30
Invalid argument in SocialAuthenticator::fetchAccessToken


ajgarlag:
Currently, the protected method `SocialAuthenticator::fetchAccessToken` requires an instance of `OAuth2Client` but it should allow any implementation of `OAuth2ClientInterface`

weaverryan:
You're totally right - the interface was added later. Can you open a PR?

Thanks!

PR #211, opened 2019-11-07
Allow any OAuth2ClientInterface implementation in ClientRegistry

ajgarlag:
The client registry should allow any OAuth2ClientInterface implementation.

Fixes and error introduced in https://github.com/knpuniversity/oauth2-client-bundle/commit/61af3d7c7510e5f51537d3895e239aec87860c98 similar to #159.


weaverryan:
Thanks @ajgarlag!

PR #219, opened 2019-12-12
Add IBM App ID provider

Jampire:


bocharsky-bw:
Hi @Jampire ! Thank you for this contribution and confgrats on your first contribution to this project! Could you fix tests here, please? I think it's as simple as run `./bin/update_readme`, commit changes and push again.

Cheers!

src/Client/Provider/AppIdClient.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return parent::fetchUserFromToken($accessToken);
+    }
+
+    /**
+     * @return AppIdResourceOwner|ResourceOwnerInterface
+     */
+    public function fetchUser()
+    {
+        return parent::fetchUser();
```
bocharsky-bw: Though I don't see a value in calling `parent::` manually in this class, we can just omit overriding :)

README.md
```diff
@@ -458,6 +459,26 @@ knpu_oauth2_client:
...
+        appid:
+            # must be "appid" - it activates that type!
+            type: appid
+            # add and configure options in your parameters.yml or env
+            client_id: '%env(OAUTH_APPID_CLIENT_ID)%'
+            client_secret: '%env(OAUTH_APPID_CLIENT_SECRET)%'
+            tenant_id: '%env(OAUTH_APPID_TENANT_ID)%'
+            base_auth_uri: '%env(OAUTH_APPID_BASE_AUTH_URI)%'
+            # a route name you'll create, for example connect_appid_check 
+            redirect_route: '%env(OAUTH_APPID_REDIRECT_ROUTE)%'
```
bocharsky-bw: This usually should be as simple as:
`redirect_route: connect_appid_check`

src/DependencyInjection/Providers/AppIdProviderConfigurator.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public function getProviderOptions(array $configuration)
+    {
+        return [
+            'clientId' => $configuration['client_id'],
+            'clientSecret' => $configuration['client_secret'],
+            'baseAuthUri' => $configuration['base_auth_uri'],
+            'tenantId' => $configuration['tenant_id'],
+            'redirectRouteName' => $configuration['redirect_route'],
```
bocharsky-bw: Let's be consistent: if we name it `redirect_route` - let's it be `redirectRoute` then

src/DependencyInjection/Providers/AppIdProviderConfigurator.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public function getProviderClass(array $configuration)
+    {
+        return AppIdProvider::class;
+    }
+
+    public function getClientClass(array $config)
+    {
+        return AppIdClient::class;
```
bocharsky-bw: This should be a "string" value, otherwise, we will have "class not found" error, but we don't want to enforce users to have `jampire/oauth2-appid` even if they don't use it. See as example: https://github.com/knpuniversity/oauth2-client-bundle/blob/9997fbec0ad46d6420e0cfd95486258c5623343e/src/DependencyInjection/Providers/FacebookProviderConfigurator.php

src/DependencyInjection/Providers/AppIdProviderConfigurator.php
```diff
@@ -0,0 +1,75 @@
...
+                ->defaultValue('saml')
+                ->info('Identity Provider code. More detail: https://cloud.ibm.com/docs/services/appid?topic=appid-getting-started')
+                ->example('idp: saml')
+            ->end()
+        ;
+    }
+
+    public function getProviderClass(array $configuration)
+    {
+        return AppIdProvider::class;
```
bocharsky-bw: The same here, this should be a sting value of FQCN

Jampire:
@bocharsky-bw , Hi. PR is updated.

Jampire:
Hi @bocharsky-bw . Is there any reasons why this PR is not merged? Do I need to fix something?

bocharsky-bw:
Hey @Jampire ! Thanks for applying more changes here!

Yes, first of all, we need to fix failed tests and make Travis green again. Looking at failed builds, I could tell you that we need to re-run `./bin/update_readme` command to generate the correct README and that's it probably. It looks like the config example was inserted in the correct place, we order configs alphabetically there

Jampire:
Hi @bocharsky-bw this is weird, because the issue was in #221 PR. I regenerated README.md. TravisCI is green again.

weaverryan:
Thanks to both of you! Released in 1.33.0

PR #221, opened 2019-12-19
Oauth apple support

lolmx:
Adds support for [oauth-apple](https://github.com/patrickbussmann/oauth2-apple)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ via Composer:
 | OAuth2 Provider                                                       | Install                                             |
 | --------------------------------------------------------------------- | ------------------------------------------------------ |
 | [Amazon](https://github.com/luchianenco/oauth2-amazon)                | composer require luchianenco/oauth2-amazon          |
+| [Apple](https://github.com/patrickbussmann/oauth2-apple)              | composer require patrickbussmann/oauth2-apple       |
 | [Auth0](https://github.com/RiskioFr/oauth2-auth0)                     | composer require riskio/oauth2-auth0                |
 | [Azure](https://github.com/thenetworg/oauth2-azure)                   | composer require thenetworg/oauth2-azure            |
 | [Bitbucket](https://github.com/stevenmaguire/oauth2-bitbucket)        | composer require stevenmaguire/oauth2-bitbucket     |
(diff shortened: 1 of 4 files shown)


src/DependencyInjection/Providers/AppleProviderConfigurator.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function getClientClass(array $config)
+    {
+        return 'KnpU\OAuth2ClientBundle\Client\Provider\AppleClient';
+    }
+
+    public function getProviderOptions(array $configuration)
+    {
+        return [
+            'clientId' => $configuration['client_id'],
```
bocharsky-bw: Ah, wait, `client_secret` is missing here :)

lolmx: apple does not have a client_secret, they use the key_file instead

bocharsky-bw: In this case, we should remove `client_secret` from the docs I think

README.md
```diff
@@ -458,6 +459,24 @@ knpu_oauth2_client:
...
 
+        # will create service: "knpu.oauth2.client.apple"
+        # an instance of: KnpU\OAuth2ClientBundle\Client\Provider\AppleClient
+        # composer require patrickbussmann/oauth2-apple
+        apple:
+            # must be "apple" - it activates that type!
+            type: apple
+            # add and configure client_id and client_secret in parameters.yml
+            client_id: '%env(OAUTH_APPLE_CLIENT_ID)%'
+            client_secret: '%env(OAUTH_APPLE_CLIENT_SECRET)%'
```
bocharsky-bw: No `client_secret`? We need to find a way to remove this line then to avoid confusion. I wonder if we have any other clients in this bundle that don't have `client_secret` as well, how do we remove this field from docs?

lolmx: at beginning i removed it and added the needed custom params, but it causes the build to fail ><
i've also tried to add a comment on the line, same build fails..

i dunno :/

weaverryan:
I'll fix up the README - my weird script for it probably needs to be tweaked :)

weaverryan:
Tagged in 1.32.0.

Thanks!

PR #226, opened 2020-02-07
Fixed minor typo

sdaoudi:
Add missing ":" for http_client_options configuration

PR #227, opened 2020-02-08
Updates to bin/update_readme script to fit current Symfony version

Jampire:
Hi. I updated bin/update_readme to fit current Symfony versions and re-generated README.md file applying these changes. Also I included PR #226 as there is no sense to update README.md file only.

### DIFF
--- a/bin/update_readme
+++ b/bin/update_readme
@@ -38,7 +38,7 @@ $sectionTemplate = <<<EOF
         %PROVIDER_NAME%:
             # must be "%PROVIDER_NAME%" - it activates that type!
             type: %PROVIDER_NAME%
-            # add and configure client_id and client_secret in parameters.yml
+            # add and configure client_id and client_secret in parameters.yml or .env* files
             client_id: '%env(OAUTH_%UCPROVIDER_NAME%_CLIENT_ID)%'
             %CLIENT_SECRET%
             # a route name you'll create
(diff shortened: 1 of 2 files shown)


bin/update_readme
```diff
@@ -38,7 +38,7 @@ $sectionTemplate = <<<EOF
         %PROVIDER_NAME%:
             # must be "%PROVIDER_NAME%" - it activates that type!
             type: %PROVIDER_NAME%
-            # add and configure client_id and client_secret in parameters.yml
+            # add and configure client_id and client_secret in parameters.yml or .env* files
```
weaverryan: Let's go full "new Symfony" with this :) 

```suggestion
            # add and set these environment variables in your .env files
```

weaverryan:
Great catch and fix for these outdated docs! Awesome! I tweaked a bit further in sha: 1483e391ef71ca5558fb1ddf9e2fec97dd248474 to only show the "new" way of doing things.

Thanks!

bin/update_readme
Jampire: Hi @weaverryan , I thought about this initially. But I noticed that `oauth2-client-bundle` supports Symfony v3.4. And this version still uses parameters.yml.

weaverryan: It does - but you're own your own as far as documentation goes :). It's impossible to write good documentation for multiple versions at once. For example, just mentioning parameters.yml here isn't enough - 3.4 doesn't use the `env()` syntax at all! Better to be clear and on the latest version

PR #230, opened 2020-03-06
Allow to pass options when fetching access token

akondas:
This little change allow to pass `options` when fetching `AccessToken`. 

### Motivation
Some providers checks for `redirect_uri` when fetching token from `code` (last step). When in my app is more than one entry point with otuah (for example different scopes in different app region) I want to have ability to change `redirect_uri`. The `redirect` method allows for that, but then, when user returns from provider `redirect_uri` param is taken from main configuration, which causes auth error (i.e. GitLab). For now the only solution is:
```
        gitlab-register:
            type: gitlab
            client_id: '%env(OAUTH_GITLAB_CLIENT_ID)%'
            client_secret: '%env(OAUTH_GITLAB_CLIENT_SECRET)%'
            redirect_route: register_gitlab_check
        gitlab-auth:
            type: gitlab
            client_id: '%env(OAUTH_GITLAB_CLIENT_ID)%'
            client_secret: '%env(OAUTH_GITLAB_CLIENT_SECRET)%'
            redirect_route: login_gitlab_check
```

This PR allow to have one client and change only `redirect_uri`:

```php
// Start -> redirect
$this->oauth->getClient('gitlab')->redirect(['read_user'], ['redirect_uri' => 'https://some.url'])

// Return -> check
$this->oauth->getClient('gitlab')->getAccessToken(['redirect_uri' => 'https://some.url']);
```

### DIFF
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -80,6 +80,8 @@ class OAuth2Client implements OAuth2ClientInterface
     /**
      * Call this after the user is redirected back to get the access token.
      *
+     * @param array $options Additional options that should be passed to the getAccessToken() of the underlying provider
+     *
      * @return AccessToken|\League\OAuth2\Client\Token\AccessTokenInterface
      *
      * @throws InvalidStateException
--- a/src/Client/OAuth2Client.php
+++ b/src/Client/OAuth2Client.php
@@ -86,7 +86,7 @@ class OAuth2Client implements OAuth2ClientInterface
      * @throws MissingAuthorizationCodeException
      * @throws IdentityProviderException         If token cannot be fetched
      */
-    public function getAccessToken()
+    public function getAccessToken(array $options = [])
     {
         if (!$this->isStateless) {
             $expectedState = $this->getSession()->get(self::OAUTH2_SESSION_STATE_KEY);
@@ -102,9 +102,10 @@ class OAuth2Client implements OAuth2ClientInterface
             throw new MissingAuthorizationCodeException('No "code" parameter was found (usually this is a query parameter)!');
         }
 
-        return $this->provider->getAccessToken('authorization_code', [
-            'code' => $code,
-        ]);
+        return $this->provider->getAccessToken(
+            'authorization_code',
+            array_merge(['code' => $code], $options)
+        );
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Client/OAuth2Client.php
```diff
@@ -15,6 +15,7 @@
 use League\OAuth2\Client\Provider\AbstractProvider;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Token\AccessToken;
+use phpDocumentor\Reflection\DocBlock\Tags\Param;
```
akondas: :man_facepalming:

src/Client/OAuth2ClientInterface.php
```diff
@@ -39,7 +39,7 @@ public function redirect(array $scopes, array $options);
      * @throws \KnpU\OAuth2ClientBundle\Exception\MissingAuthorizationCodeException
      * @throws \League\OAuth2\Client\Provider\Exception\IdentityProviderException   If token cannot be fetched
      */
-    public function getAccessToken();
+    public function getAccessToken(array $options = []);
```
weaverryan: We should describe this an `@param` - something like:

> Additional options that should be passed to the getAccessToken() of the underlying provider

akondas: Ready :wink:

akondas:
One question: adding new optional param to method definition is really a bc-break?

weaverryan:
Yea... unfortunately it is - you'll get stuff like this for classes that already implement it:

> PHP Fatal error:  Declaration of FooBar::getAccessToken() must be compatible with KnpU\OAuth2ClientBundle\Client\OAuth2ClientInterface::getAccessToken(array $options = Array) in /Users/weaverryan/Sites/os/KnpUOAuth2ClientBundle/play.php on line 20

And we mention it in the Symfony BC details here - https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces - even for me, I need to check this to be sure - BC is super complicated.

I'll review and hopefully merge this soon :)

weaverryan:
Thank you @akondas!

I'm going to wait for #232 before merging - because I want to make sure that doesn't impact this.

Cheers!

PR #237, opened 2020-04-23
List required symfony version

psociety:
I have symfony 4.3, so can't install this package and i only saw there was a minimum version during composer install :(


weaverryan:
Thanks @psociety! I also added to the note that you CAN use v1 of this bundle for earlier versions of Symfony.

Cheers!

Issue #253, opened 2020-07-20
Sign in with Apple - Authentication failed! Did you authorize our app?


bogdandovgopol:
Hello,

I'm trying to implement sing in with apple, but I'm getting this error `Authentication failed! Did you authorize our app?` when I'm redirected back from the Apple Sign in page.

`$this->clientRegistry->getClient('apple')->getAccessToken();` returns `No "code" parameter was found (usually this is a query parameter)!`
It seems like Apple redirects back to my website without any parameters.

In the apple developer dashboard I created App ID, Service ID(Which is the client_id) and generated a new key. 

Checked multiple times to make sure that client_id, team_id and key_file_id are correct...

Also have the GoogleAuthenticator which works just fine.

Here is my config: 
```
        apple:
            # must be "apple" - it activates that type!
            type: apple
            # add and set these environment variables in your .env files
            client_id: '%env(OAUTH_APPLE_CLIENT_ID)%'
            # a route name you'll create
            redirect_route: connect_apple_check
            redirect_params: {}
            team_id: [TEAM_ID]
            key_file_id: [KEY_FILE_ID]
            key_file_path: '/var/www/vhosts/domain.com/httpdocs/config/keys/apple/AuthKey_[KEY_FILE_ID].p8'
            # whether to check OAuth2 "state": defaults to true
            use_state: false 
```

AppleAuthenticator.php
```
class AppleAuthenticator extends SocialAuthenticator
{

    private ClientRegistry $clientRegistry;
    private EntityManagerInterface $em;
    private UserPasswordEncoderInterface $passwordEncoder;
    private UrlGeneratorInterface $urlGenerator;
    private EventDispatcherInterface $eventDispatcher;
    private SessionInterface $session;

    public function __construct(ClientRegistry $clientRegistry, EntityManagerInterface $em, UserPasswordEncoderInterface $passwordEncoder, UrlGeneratorInterface $urlGenerator, EventDispatcherInterface $dispatcher, SessionInterface $session)
    {
        $this->clientRegistry = $clientRegistry;
        $this->em = $em;
        $this->passwordEncoder = $passwordEncoder;
        $this->urlGenerator = $urlGenerator;
        $this->eventDispatcher = $dispatcher;
        $this->session = $session;
    }

    public function supports(Request $request)
    {
        // continue ONLY if the current ROUTE matches the check ROUTE
        return $request->attributes->get('_route') === 'connect_apple_check';
    }

[...]

Issue #249, opened 2020-06-25
Apple Login sends "state" as POST instead of GET


Dandrum:
We get the Error `Invalid state parameter passed in callback URL.` Because the state isn't a GET value. So far i know they changed it because of security issues or so. Is it possible to change if state check is GET or POST for a provider ?

Best,
Dan

weaverryan:
Hey @Dandrum!

First, in case it helps, here is where the "state" check is performed: https://github.com/knpuniversity/oauth2-client-bundle/blob/2e54646ae5b1a41bde7985ab1dbdd5cb4ceb0462/src/Client/OAuth2Client.php#L93-L99


> We get the Error Invalid state parameter passed in callback URL. Because the state isn't a GET value

I don't know anything about Apple Login, but when they redirect back to your site, I don't think they can do that with anything other than a GET request (but I could be wrong). Anyways, if that is the case (that the state is POST data), you'll probably need to set your client as "stateless"  `use_state: false` and then check the state manually yourself (including setting the state to the session before redirecting to Apple). It's not too much work - so I think that might be the simplest.

I'm not against also looking for state as a POST parameter, but I've never heard of this before with OAuth.

Cheers!

Dandrum:
Hi,
i think it was an security issue so apple changed it to POST instead of GET. 
The apple sign in bundle you ref to in your docs are using POST too -> [https://github.com/patrickbussmann/oauth2-apple#authorization-code-flow ](https://github.com/patrickbussmann/oauth2-apple#authorization-code-flow )

And the solution in apple docs: -> [https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/configuring_your_webpage_for_sign_in_with_apple#3331292](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/configuring_your_webpage_for_sign_in_with_apple#3331292)

Hope that helps you a bit.
Best, Dan

weaverryan:
Thanks for the info @Dandrum!

Indeed, there seems to be a `form_post` OAuth mode that I wasn't familiar with - https://auth0.com/docs/protocols/oauth2 describes it. And so, our state checking should check for state in both `$request->query()` and `$request->request()`. I'd welcome a pull request if you'd like to make that change. Probably we should check `$request->request()` first (i.e. give that one priority).

PR #254, opened 2020-07-21
Look for both POST and GET params when getting state

bogdandovgopol:
#253 #249


weaverryan:
Thank you @bogdandovgopol! Tagged in 2.2.0

PR #265, opened 2020-09-27
Improve test coverage

foxworth42:
While investigating an edge case I have where the redirect_uri isn't being generated properly (still doing some troubleshooting before I raise an issue) I started looking through the tests for the package and saw there were some chunks that weren't under test.  I figured I would spend some time writing additional tests for exercise.


weaverryan:
This is awesome! You're totally right - I was a bit surprised (even though I wrote the original library) about some of these missing pieces. You rock! Thank you :)

PR #266, opened 2020-09-28
Add specific error message.

### LINKED ISSUE #222: Wrong generic provider class error is not explicit
I made a mistake with the `provider_class` options of my generic configuration:

```yaml
knpu_oauth2_client:
    clients:
        fusionauth:
            type: generic
            # The class does not exist, I only put the namespace.
            provider_class: League\OAuth2\Client\Provider
            provider_options:
                urlAuthorize: https://fusionauth.lab.nexylan.dev/oauth2/authorize
                urlAccessToken: https://fusionauth.lab.nexylan.dev/oauth2/token
                urlResourceOwnerDetails: https://fusionauth.lab.nexylan.dev/oauth2/userinfo
            client_id: "%env(resolve:FUSIONAUTH_CLIENT_ID)%"
            client_secret: "%env(resolve:FUSIONAUTH_CLIENT_SECRET)%"
            redirect_route: home
```

So it's a pebkac, of course, but the produced error is not relevant at all:

```
Run `composer require ` in order to use the "generic" OAuth provider.
```

You may add a previous message telling the class is not found before this suggestion. With the current message, I didn't realized my mistake at the first time. I would if I had a message like "The provider class <CLASS> does not exist.".

foxworth42:
Add specific message when a generic client has configured a provider_class that does not exist.

resolves #222

### DIFF
--- a/src/DependencyInjection/KnpUOAuth2ClientExtension.php
+++ b/src/DependencyInjection/KnpUOAuth2ClientExtension.php
@@ -245,6 +245,9 @@ class KnpUOAuth2ClientExtension extends Extension
     private function configureProviderAndClient(ContainerBuilder $container, $providerType, $providerKey, $providerClass, $clientClass, $packageName, array $options, $redirectRoute, array $redirectParams, $useState, array $collaborators)
     {
         if ($this->checkExternalClassExistence && !class_exists($providerClass)) {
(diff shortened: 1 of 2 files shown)


weaverryan:
Beautiful - thanks!

PR #268, opened 2020-10-09
Add Okta client

foxworth42:
Finally following up officially on adding a client for integrating with Okta (been using it for quite a while as a generic provider).  Also added a 'test' command to composer to make running of tests & phpstan locally a little easier.


weaverryan:
Awesome, thank you @foxworth42!

Issue #260, opened 2020-09-02
How to check if access token expired, to be refreshed?


kdefives:
Hello,

I get the user access token from external OAuth2 system during the authentication but the access_token is valid during 2 hours.
**How should I have to check if the user access token is expired or not in order to trig a token refresh?** I do not find documentations about that part :(

I need to check if the token is valid regularly because inside my application I need to do API call to other application with this access token in header. That is why I need to be sure the token is still valid when I do the call .

For information I am using Generic type and I develop my custom provider. I am using Guard too.

**The expiration check and refresh token management is not fully managed by know/OAuth2 ?**

Thx in advance for your help

kdefives:
Hello @weaverryan ,
Does knpuniverty/oauth2-client-bundle automatically manage and detect if the access token have to be refreshed? In that case, if token need to be refresh, we could imagine knpuniverty/oauth2-client-bundle could refresh the token automatically?

Or do I have to manage that part myself, on my application ? For instance using Events / EventsListeners using Symfony?

Thx

ndench:
Hey @kdefives I have the same issue where I want an easy way to refresh the access token.

This bundle just uses league/oauth2-client under the hood, so you can see how to refresh a token from [their documentation](https://github.com/thephpleague/oauth2-client/blob/419d163a02654305773e07731dc26ce7528a574f/docs/usage.md#refreshing-a-token). To refresh within the context of this bundle, you can do:

```php
$provider = $clientRegistry->getClient('your client name')->getOAuth2Provider();

$existingAccessToken = getAccessTokenFromYourDataStore();

if ($existingAccessToken->hasExpired()) {
    $newAccessToken = $provider->getAccessToken('refresh_token', [
        'refresh_token' => $existingAccessToken->getRefreshToken()
    ]);

    // Purge old access token and store new access token to your data store.
}
```

IMO I think this bundle should natively support refreshing the token. The reason that it doesn't is because the `authorization_code` grant type is hard coded into [OAuth2Client::getAccessToken](https://github.com/knpuniversity/oauth2-client-bundle/blob/73e8f09d733309726df7f6f00e18b4522a6a499f/src/Client/OAuth2Client.php#L110).

I'm not sure on the best way to add this feature, maybe @weaverryan could comment? I have a couple of ideas:
1. Make `grantType` a parameter to the `getAccessToken` method (it would have to be an optional second parameter with a default value to protect BC).
2. Make a new `refreshToken` method.

[...]

PR #277, opened 2020-11-12
Add support for refreshToken()

ndench:
As discussed in #260, it would be great to have native support to refresh access tokens. I'm successfully using this in my Symfony 4.4 application.

Closes: #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ If you store your access tokens for use at a later time, you should check that i
 If it is expired, you can easily refresh it:
 
 ```php
-// Load the access token from wherever you have store it (eg. user session, database, etc)
+// Load the access token from persistence (eg. user session, database, etc)
 $accessToken = $tokenStorage->retrieveAccessToken();
 
 if ($accessToken->hasExpired()) {
--- a/README.md
+++ b/README.md
@@ -430,6 +430,29 @@ public function getUser($credentials, UserProviderInterface $userProvider)
 The logged-in user will be an instance of `KnpU\OAuth2ClientBundle\Security\User\OAuthUser` and will
 have the roles `ROLE_USER` and `ROLE_OAUTH_USER`.
 
+## Refresh Tokens
+
+If you store your access tokens for use at a later time, you should check that it's not expired before you use it again.
+If it is expired, you can easily refresh it:
+
+```php
+// Load the access token from wherever you have store it (eg. user session, database, etc)
+$accessToken = $tokenStorage->retrieveAccessToken();
+
+if ($accessToken->hasExpired()) {
+    $accessToken = $client->refreshAccessToken($accessToken);
+
+    // Update the stored access token for next time
+    $tokenStorage->updateAccessToken($accesStoken);
+}
+```
+
+Depending on your OAuth2 provider, you may need to pass some parameters when refreshing the token:
+
+```php
+$accessToken = $client->refreshAccessToken($accessToken, ['scopes' => 'offline_access']);
+```
+
 ## Configuration
 
 Below is the configuration for *all* of the supported OAuth2 providers.
--- a/README.md
+++ b/README.md
@@ -430,21 +430,38 @@ public function getUser($credentials, UserProviderInterface $userProvider)
 The logged-in user will be an instance of `KnpU\OAuth2ClientBundle\Security\User\OAuthUser` and will
 have the roles `ROLE_USER` and `ROLE_OAUTH_USER`.
 
-## Refresh Tokens
+## Storing and refreshing tokens
 
-If you store your access tokens for use at a later time, you should check that it's not expired before you use it again.
-If it is expired, you can easily refresh it:
+You have a couple of options to store access tokens for use at a later time:
 
-```php
-// Load the access token from persistence (eg. user session, database, etc)
-$accessToken = $tokenStorage->retrieveAccessToken();
+1. Store the `AccessToken` object (eg. serializing into the session), this allows you to check expiry before refreshing
+    ```php
+    // Fetch and store the AccessToken
+    $accessToken = $client->getAccessToken();
+    $session->set('access_token', $accessToken);
 
-if ($accessToken->hasExpired()) {
-    $accessToken = $client->refreshAccessToken($accessToken);
+    // Load the access token from the session, and refresh if required
+    $accessToken = $session->get('access_token');
 
-    // Update the stored access token for next time
-    $tokenStorage->updateAccessToken($accesStoken);
-}
+    if ($accessToken->hasExpired()) {
+        $accessToken = $client->refreshAccessToken($accessToken);
+
+        // Update the stored access token for next time
+        $session->set('access_token', $accessToken);
+    }
+    ```
+
+2. Store just the refresh token string (eg. in the dabatase `user.refresh_token`), this means you must always refresh
+    ```php
+    // Fetch the AccessToken and store the refresh token
+    $accessToken = $client->getAccessToken();
+    $user->setRefreshToken($accessToken->getRefreshToken());
+    $entityManager->flush();
+
+    // Get a new AccessToken from the refresh token, and store the new refresh token for next time
+    $accessToken = $client->refreshAccessToken($user->getRefreshToken);
+    $user->setRefreshToken($accessToken->getRefreshToken());
+    $entityManager->flush();
 ```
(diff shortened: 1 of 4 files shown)


src/Client/OAuth2ClientInterface.php
```diff
@@ -45,6 +47,17 @@ public function redirect(array $scopes, array $options);
...
+    /**
+     * Refresh the given AccessToken.
+     *
+     * @param array $options Additional options that should be passed to the getAccessToken() of the underlying provider
+     *
+     * @return AccessToken|\League\OAuth2\Client\Token\AccessTokenInterface
+     *
+     * @throws IdentityProviderException If token cannot be fetched
+     */
+    public function refreshAccessToken(AccessTokenInterface $accessToken, array $options = []);
```
weaverryan, line 59: For backwards-compatibility (so that we don't need to bump a new major version) - could you *not* actually add the method to the interface, but instead add it as documentation above the method? Here is an example: https://github.com/symfony/symfony/pull/36515/files#diff-a1c8b41afc0729b212c093eba2d2e82fdd13e22884fbc5fdf7575f3ad1fab0c6

I'm not sure if we can document the exact `$options` parameter with as much detail as we have here, so we may need to move that into the description of the method in that new format.

weaverryan (commented):
Sorry for the delay! This is wonderful! Just 2 things:

1) I added a note for a change so we don't break backwards-compatibility.
2) Can you add something to the README.md? Maybe a new, small section called "Refresh Tokens" Before `Configuration` that shows how to use this new method, and maybe some details about how you probably need to ask for a refresh token (e.g. via a `offline_access` scope).

Thanks!

src/Client/OAuth2ClientInterface.php
ndench: Yeah no problem, I'm curious how adding to an interface breaks BC though? I thought adding things was fine, it's only when you move/change existing things that BC break?

ndench (commented):
No problem @weaverryan, thanks for the feedback. I'll update the README and the interface. Just a couple questions to clarify:

1. Left a question on the Interface about the BC break, this is just personal curiosity
2. Is passing a scope to the `getAccessToken()` method as simple as `['scope' => ['offline_access']]`? Looking through the code the only place we directly reference scope is in the `redirect()` method.

src/Client/OAuth2ClientInterface.php
weaverryan: Very fair question :). If we supply an interface in this bundle, then we have to assume (even if it's not a normal use-case) that someone who uses this bundle may have created their own custom implementation (class) for this interface. If we add a new method in a minor release, suddenly their class explodes because it's missing 1 method from the interface. It's total nerdery, but you can go to http://symfony.com/bc to read about all the rules of what is and is not allowed for the purposes of backwards-compatibility in Symfony's codebase :).

weaverryan:
> Is passing a scope to the getAccessToken() method as simple as ['scope' => ['offline_access']]? Looking through the code the only place we directly reference scope is in the redirect() method.

I think so? I can't remember if the name of this scope is standard or dependent on the OAuth2 server. But yes, that's more or less my impression. And yes, the user would pass this to the `->redirect()` function if they need this scope.

ndench:
@weaverryan sorry for taking so long to fix this up and thanks for explaining the BC break, I certainly hadn't considered that before!

Let me know if there's anything else that needs to be fixed up.

README.md
```diff
@@ -430,6 +430,29 @@ public function getUser($credentials, UserProviderInterface $userProvider)
 The logged-in user will be an instance of `KnpU\OAuth2ClientBundle\Security\User\OAuthUser` and will
 have the roles `ROLE_USER` and `ROLE_OAUTH_USER`.
 
+## Refresh Tokens
+
+If you store your access tokens for use at a later time, you should check that it's not expired before you use it again.
+If it is expired, you can easily refresh it:
+
+```php
+// Load the access token from wherever you have store it (eg. user session, database, etc)
```
jrushlow, line 439:
```suggestion
// Load the access token from persistence (eg. user session, database, etc)
```

jrushlow: Looking good!

ndench:
Thanks @jrushlow, I've applied your suggestion. That's a much better wording.

jrushlow (approved):

src/Client/OAuth2Client.php
```diff
@@ -112,6 +113,23 @@ public function getAccessToken(array $options = [])
...
+    /**
+     * Refresh the given AccessToken.
+     *
+     * @param array $options Additional options that should be passed to the getAccessToken() of the underlying provider
+     *
+     * @return AccessToken|\League\OAuth2\Client\Token\AccessTokenInterface
+     *
+     * @throws IdentityProviderException If token cannot be fetched
+     */
+    public function refreshAccessToken(AccessTokenInterface $accessToken, array $options = [])
```
weaverryan, line 125: Something just occurred to me. Let's think of the flow:

1) After the OAuth flow, I call `getAccessToken()` and get back the `AccessToken` object.
2) I decide to store this in the database - on an `User.accessToken` property
3) Now... I no longer have the `AccessToken` object, I only have the access token string. So, I can't use `$accessToken->hasExpired()` (like the new docs show) to see if the token has expired. But more importantly, I can't call this method because I don't have the object anymore. 

WDYT? Am I missing something? Would you recommend a different strategy for storing the access token object? Or should this method's first argument be `string $refreshToken` and then we document how you should store the refresh token string in (e.g.) the database?

ndench: That's a good point. The alternative (and what I'm currently using) is to store the whole AccessToken (ie. token string, refresh token and expiry), so that you can check if it's expired and refresh if required.

However, just storing the refresh token and refreshing everytime you want to use it is a more simple flow.

I think making this method take the `string $refreshToken` is the most flexible, it allows both flows.

README.md
```diff
@@ -430,6 +430,46 @@ public function getUser($credentials, UserProviderInterface $userProvider)
...
+    ```php
+    // Fetch and store the AccessToken
+    $accessToken = $client->getAccessToken();
+    $session->set('access_token', $accessToken);
+
+    // Load the access token from the session, and refresh if required
+    $accessToken = $session->get('access_token');
+
+    if ($accessToken->hasExpired()) {
+        $accessToken = $client->refreshAccessToken($accessToken);
```
weaverryan, line 447: needs to be `$accessToken->getRefreshToken()`, right?

README.md
```diff
@@ -430,6 +430,46 @@ public function getUser($credentials, UserProviderInterface $userProvider)
...
+    // Get a new AccessToken from the refresh token, and store the new refresh token for next time
+    $accessToken = $client->refreshAccessToken($user->getRefreshToken);
+    $user->setRefreshToken($accessToken->getRefreshToken());
+    $entityManager->flush();
+```
+
+Depending on your OAuth2 provider, you may need to pass some parameters when refreshing the token:
+
+```php
+$accessToken = $client->refreshAccessToken($accessToken, ['scopes' => 'offline_access']);
```
weaverryan, line 470: I think they may also need to do this during the original redirect.

README.md
```diff
@@ -430,6 +430,46 @@ public function getUser($credentials, UserProviderInterface $userProvider)
...
+        $session->set('access_token', $accessToken);
+    }
+    ```
+
+2. Store just the refresh token string (eg. in the dabatase `user.refresh_token`), this means you must always refresh
+    ```php
+    // Fetch the AccessToken and store the refresh token
+    $accessToken = $client->getAccessToken();
+    $user->setRefreshToken($accessToken->getRefreshToken());
+    $entityManager->flush();
```
weaverryan, line 459: You could ALSO store the `access_token` and expires in the database if you wanted to. Maybe we could make a note of that?

```php
// you can also store the access token and expiration and then avoid the refresh
// until the access token is actually expired
```

weaverryan (commented):
This is looking good - just a few minor documentation requests. Thank you @ndench

README.md
ndench: Good catch!

ndench (commented):
Thanks for the feedback, everyone loves thorough documentation!

weaverryan:
Thanks @ndench for your quality, consistent work on this! This is now part of 2.7.0 :)

PR #280, opened 2020-11-23
Allow the installation on php 8

### LINKED ISSUE #275: Support for PHP 8
With PHP 8 around the corner, it would be nice to support this version as well. Even if it's RC, I'm checking the compatibility of this version in my project.

derrabus:
Hello! This PR proposes to allow the installation of this package on php 8 because it seems to work. It also extends the Travis configuration to make sure that the project is tested on php 8.

Fixes #275

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ cache:
 php:
     - 7.1
     - 7.2
-    - 7.3
+    - 7.4
+    - 8.0snapshot
 matrix:
     include:
         - php: 7.3
@@ -20,6 +21,7 @@ matrix:
 before_install:
     - phpenv config-rm xdebug.ini || true
     - composer self-update
+    - if [ "$LINT" != 1 ]; then composer remove --dev --no-update phpstan/phpstan; fi;
 
 install: composer update --prefer-dist $COMPOSER_FLAGS
 
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.1.3|^8.0",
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -9,7 +9,8 @@ cache:
 php:
     - 7.1
     - 7.2
-    - 7.3
```
derrabus, line 12: 💡 I've removed the 7.3 run here because `matrix.include` already contains a run on PHP 7.3, so this one was redundant.

.travis.yml
```diff
@@ -20,6 +21,7 @@ matrix:
 before_install:
     - phpenv config-rm xdebug.ini || true
     - composer self-update
+    - if [ "$LINT" != 1 ]; then composer remove --dev --no-update phpstan/phpstan; fi;
```
derrabus, line 24: 💡 This Projects locks phpstan at version 0.11 which is not compatible with php 8. You probably should bump to 0.12 eventually, but this seemed out of scope for this PR. This is why I chose to remove phpstan when it's not needed.

composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.1.3|^8.0",
```
derrabus, line 15: For the record, my personal preference would be the following constraint.

```suggestion
        "php": ">=7.1.3",
```

But I'd like to leave it up to the maintainers to define the strategy here.

weaverryan:
Thank you @derrabus for this nice, clean PR! There maybe a PHPUnit failure in PHP 7.1... but we'll see if it happens again. It's odd - an undefined method that should be there...

PR #300, opened 2021-04-16
Adding FusionAuth provider.

mooreds:
Modelled this after keycloak provider. Added docs and code.

### DIFF
--- a/src/DependencyInjection/Providers/FusionAuthProviderConfigurator.php
+++ b/src/DependencyInjection/Providers/FusionAuthProviderConfigurator.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * OAuth2 Client Bundle
+ * Copyright (c) KnpUniversity <http://knpuniversity.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+
+class FusionAuthProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        $node
+            ->scalarNode('auth_server_url')
+                ->isRequired()
+                ->info('FusionAuth Server URL, no trailing slash')
+            ->end()
+        ;
+    }
+
+    public function getProviderClass(array $config)
+    {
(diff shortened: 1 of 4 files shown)


src/DependencyInjection/Providers/FusionAuthProviderConfigurator.php
```diff
@@ -0,0 +1,64 @@
...
+            ->scalarNode('auth_server_url')
+                ->isRequired()
+                ->info('FusionAuth Server URL, no trailing slash')
+            ->end()
+        ;
+    }
+
+    public function getProviderClass(array $config)
+    {
+        return 'JerryHopper\OAuth2\Client\Provider\FusionAuth';
```
weaverryan, line 29: We may not do this consistently everywhere, but could you add a `use` statement for this and then use `FusionAuth::class` here?

src/DependencyInjection/Providers/FusionAuthProviderConfigurator.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function getProviderDisplayName()
+    {
+        return 'FusionAuth';
+    }
+
+    public function getClientClass(array $config)
+    {
+        return 'KnpU\OAuth2ClientBundle\Client\Provider\FusionAuthClient';
```
weaverryan, line 62: Same here - `use` statement and then `::class`

weaverryan (commented):
Looks good  - just some minor changes needed

mooreds:
Changed those methods to return `::class` objects. Thanks for the feedback!

src/DependencyInjection/Providers/FusionAuthProviderConfigurator.php
bocharsky-bw: @weaverryan Wait, won't this cause a "class not found" error when someone does not have this FusionAuth provider installed? IIRC we use it as string so that PHP won't parse it, but with `FusionAuth::class` syntax it will look for this class and if it's not installed - throw an error

mooreds: Please let me know how you'd like to proceed, either way is fine with me :) .

weaverryan: No, it's ok to use `FusionAuth::class` if the class doesn't exist: PHP just finds the use statement, resolves it, then uses that class name, regardless of if it exists or not. Here's a little demo :) https://3v4l.org/oXhMl

weaverryan:
Thanks Dan!

PR #292, opened 2021-03-13
Added new Oauth2Authenticator using the new Symony 5.2 Authenticator …

Anthodev:
…interface

First draft for the implementation of the new authenticator interface, not sure that's really usable at the moment.

PR #302, opened 2021-04-26
Fixing access token, which becomes a string

weaverryan:
Renamed $credentials -> $accessToken for clarity.

Also, use $accessToken from the original authenticate() method, instead of the argument that's passed to the callback, because that will have been cast into just the access token string.

See: https://github.com/knpuniversity/oauth2-client-bundle/pull/292#issuecomment-825110798

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,12 +473,13 @@ class MyFacebookAuthenticator extends OAuth2Authenticator
 
     public function authenticate(Request $request): PassportInterface
     {
-        $accessToken = $this->fetchAccessToken($this->clientRegistry->getClient('facebook_main'));
+        $client = $this->clientRegistry->getClient('facebook_main');
+        $accessToken = $this->fetchAccessToken($client);
 
         return new SelfValidatingPassport(
-            new UserBadge($accessToken, function() {
+            new UserBadge($accessToken, function() use ($accessToken, $client) {
                 /** @var FacebookUser $facebookUser */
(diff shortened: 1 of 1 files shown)


fliespl:
@weaverryan actually what you did is just changing variable names - it doesn't fix the main issue :)

I had no better idea how to get around, so what I did in my case is following (notice $accessToken & $client passed in "use"):

```
        $accessToken = $this->fetchAccessToken($client);

        return new SelfValidatingPassport(
            new UserBadge($accessToken->getToken(), function ($notUsedTokenCauseWePassObject) use ($client, $accessToken) {
                // fetch user
                $externalUser = $client->fetchUserFromToken($accessToken);

                // create local user
                $user = new User();
                $user->setExternalId($externalUser->getId());
                $user->setType('xxx');
                $user->setAccessToken($accessToken->jsonSerialize()); // whole access token object
                // persist & flush

                return $user;
            })
        );
```

For you example to work, it would need to be like:

```
$facebookUser = $this->clientRegistry->getClient('facebook_main')->fetchUserFromToken(new AccessToken(['access_token' => $accessToken]));
```

weaverryan:
Ah, thank you @fliespl! Your strategy was exactly what I was trying to do, but I forgot the `use` (so it would not have worked). I also like how you created a `$client` variable so you can re-use it. I've just adapted the PR :). Let me know how it looks to you.

fliespl:
Cool! That will work now :)

weaverryan:
Sweet! Thanks!

PR #309, opened 2021-05-24
Update TwitchClient configuration

chypriote:
The tpavlek/oauth2-twitch provider returns a [TwitchUser](https://github.com/tpavlek/oauth2-twitch/blob/master/src/Entity/TwitchUser.php) entity.

### DIFF
--- a/src/Client/Provider/TwitchClient.php
+++ b/src/Client/Provider/TwitchClient.php
@@ -10,14 +10,14 @@
 
 namespace KnpU\OAuth2ClientBundle\Client\Provider;
 
-use AdamPaterson\OAuth2\Client\Provider\TwitchResourceOwner;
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks @chypriote! What's odd is that their TwitchUser is *supposed* to implement that ResourceOwnerInterface - you can see it in the main library here: https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L763 - but to be accurate to how the library actually works, you're totally correct.

chypriote:
Yeah, the current implementation is lacking the ResourceOwnerInterface. I actually started working on an [updated fork](https://github.com/SpireGG/oauth2-twitch) which both implements the interface and also uses the new Twitch api (helix). Let me know if you wish me to open a PR for this, as the previous client doesn't seem to be getting any updates.

PR #311, opened 2021-05-25
Rename Tests/Security/Authenticator/OAuth2AuthenticatorTest.php to tests/

### LINKED ISSUE #310: Cannot extract this bundle on NTFS filesystems from Zip archive
Version: Tag 2.8.0

Error: The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems): ZipArchive::extractTo(): File exists

This occurs because there are two directories named "Tests" and "tests" in the root folder. Users installing this bundle on an NTFS file system will get the above error.

weaverryan:
Fixes #310


erkens:
Is it possible to tell when this fix will be released? Because when using Docker for Windows with a NTFS binding it isn't possible to install with composer because of this issue.

Issue #318, opened 2021-06-23
PHP 8 Support


Matts:
Heyo
Currently the composer.json sticks to "php": ">=7.1.3"
We should see if we can make this ^7.1|^8.0

jrushlow:
Thanks again for the PR - copy of the the comment I left there:

Howdy @Matts, thanks for this addition. Unfortunately it is not needed as the existing version constraint allows any PHP version greater than `7.1.3`.  e.g. PHP 8 || PHP 9, etc.. Checkout https://getcomposer.org/doc/articles/versions.md#version-range

Symfony switched over to using the `>=` version constraint for PHP a while back and we are following suite here as well.

PR #280 implemented PHP 8 support.

Thanks again for your effort, it is greatly appreciated!

PR #319, opened 2021-07-01
Allowing PHP 8 within composer constraints

Matts:
Solves #318 

No direct dependency conflicts noticed while installing. Also cannot find any non strict checks ` == ` that are actually used in code. So I trust we should not have issues there.

Yes oauth2-client-bundle may still need a rewrite to completely go to php 8 but this is a start for projects that want to update :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1.3",
+        "php": ">=7.1.3|^8.0",
         "symfony/framework-bundle": "^4.4|^5.0",
         "symfony/dependency-injection": "^4.4|^5.0",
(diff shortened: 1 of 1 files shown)


jrushlow:
Howdy @Matts, thanks for this addition. Unfortunately it is not needed as the existing version constraint allows any PHP version greater than `7.1.3`.  e.g. PHP 8 || PHP 9, etc.. Checkout https://getcomposer.org/doc/articles/versions.md#version-range

Symfony switched over to using the `>=` version constraint for PHP a while back and we are following suite here as well.

PR #280 implemented PHP 8 support.

Thanks again for your effort, it is greatly appreciated!

Matts:
Ah then I must have been using an outdated version :)

PR #320, opened 2021-07-03
Fixing bug in docs giving AccessToken instead of string

jbonnier:
Using the code in the README throws an error.

> Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge::__construct(): Argument #1 ($userIdentifier) must be of type string, League\OAuth2\Client\Token\AccessToken given, called in /home/app/web/src/Security/GoogleAuthenticator.php on line 89

This fixes the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ class MyFacebookAuthenticator extends OAuth2Authenticator
         $accessToken = $this->fetchAccessToken($client);
 
         return new SelfValidatingPassport(
-            new UserBadge($accessToken, function() use ($accessToken, $client) {
+            new UserBadge($accessToken->getToken(), function() use ($accessToken, $client) {
                 /** @var FacebookUser $facebookUser */
                 $facebookUser = $client->fetchUserFromToken($accessToken);
 


bocharsky-bw (approved):
Hm, `AccessToken` has `__toString()` method that returns a string token: https://github.com/thephpleague/oauth2-client/blob/c1e2d288a10f28aeacd47b2d9e82e6b85fff524a/src/Token/AccessToken.php#L215 , but probably better to call it explicitly to avoid such errors. Looks good to me 👍

jbonnier:
> Hm, `AccessToken` has `__toString()` method that returns a string token: https://github.com/thephpleague/oauth2-client/blob/c1e2d288a10f28aeacd47b2d9e82e6b85fff524a/src/Token/AccessToken.php#L215 , but probably better to call it explicitly to avoid such errors. Looks good to me 👍

With the Google provider I had an error, I know the example is for Facebook, but yeah, I think it makes it clearer.

jrushlow (approved):
@jbonnier thank you for taking time to submit this!

jrushlow:
failing test related to #328

PR #327, opened 2021-07-31
Add missing use in doc

cedriclombardot:


bocharsky-bw (approved):
Nice catch, thank you!

PR #330, opened 2021-08-27
add symfony 6 return types

jrushlow:


src/KnpUOAuth2ClientBundle.php
```diff
@@ -11,16 +11,15 @@
...
 
 class KnpUOAuth2ClientBundle extends Bundle
 {
     /**
      * Overridden to allow for the custom extension alias.
-     *
-     * @return KnpUOAuth2ClientExtension
      */
-    public function getContainerExtension()
+    public function getContainerExtension(): ?ExtensionInterface
```
bocharsky-bw, line 22: `ExtensionInterface` is a different from that we had in `@return KnpUOAuth2ClientExtension` - we had incorrect return type? Or if that `KnpUOAuth2ClientExtension` still relevant, I'd recommend keeping that PHPDoc like this:
`@return KnpUOAuth2ClientExtension|ExtensionInterface`

weaverryan: It's minor - but I agree :). But I'm going to merge this

weaverryan:
Thanks Jesse!

Issue #323, opened 2021-07-08
Symfony 5.3 remove deprecations


deoomen:
Hello!

With new Symfony version 5.3 comes some new deprecations. 5.3 was released at 31.05 so it was over a month ago. Maybe this is a good moment to clear those deprecation notices. Some projects could be blocked of upgrading to 5.3 becaouse of strict test rules that do not allow to have any deprecations.

Example one:

`Class "KnpU\OAuth2ClientBundle\Security\User\OAuthUser" should implement method "Symfony\Component\Security\Core\User\UserInterface::getUserIdentifier()": returns the identifier for this user (e.g. its username or e-mailaddress)`

DonCallisto:
Whereas it's true that Symfony 5.3 is out there since months, it's also true that Symfony 4.4 is still supported (and will be for a "long" time). Unless this bundle stop prematurely that support, I think that would be kind of a trouble to get rid of deprecations (I'm not aware of the real nature of those; deprecation reported by you is just an addition, so no big issue, but others could be more sneaky).
Maybe we can run a build with Symfony 5.3 and check if any deprecation araise (IIRC during unit tests them should pop out)?

deoomen:
You're right. I forgot about compatibility with Symfony 4.4...

DonCallisto:
We can try to give it a shot; sometimes even two major versions can coexist. We don't rely upon too much Symfony components so checking the changelog for those, or run some test, should not be a big issue.

masonen:
just checked the security core. 

Those specific deprecations are only throw if the `::getUserIdentifier()` method does not exist in the user class. 
like so:

```php
// @deprecated since 5.3
            if ($user = $result->getUser() instanceof UserInterface && !method_exists($result->getUser(), 'getUserIdentifier')) {
                trigger_deprecation('symfony/security-core', '5.3', 'Not implementing method "getUserIdentifier(): string" in user class "%s" is deprecated. This method will replace "getUsername()" in Symfony 6.0.', get_debug_type($result->getUser()));
            }
```

or 

```php
// @deprecated since 5.3, change to $this->userProvider->loadUserByIdentifier() in 6.0
            if (method_exists($this->userProvider, 'loadUserByIdentifier')) {
                $user = $this->userProvider->loadUserByIdentifier($userIdentifier);
            } else {
[...]

PR #335, opened 2021-10-05
Removed some deprecations introduced in Symfony 5.*

DonCallisto:
With this, we should get rid of deprecations introduced with symfony 5.*
I've checked all the changelogs and latest test https://github.com/knpuniversity/oauth2-client-bundle/runs/3484375665
We should be compliant.

### DIFF
--- a/src/Security/User/OAuthUser.php
+++ b/src/Security/User/OAuthUser.php
@@ -38,6 +38,14 @@ class OAuthUser implements UserInterface
         return null;
     }
 
+    public function getUserIdentifier(): string
+    {
+        return $this->username;
+    }
+
+    /**
+     * @deprecated use getUserIdentifier instead
+     */
     public function getUsername(): string
     {
         return $this->username;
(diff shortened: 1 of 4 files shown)


DonCallisto:
Should resolve https://github.com/knpuniversity/oauth2-client-bundle/issues/323

src/Security/User/OAuthUser.php
```diff
@@ -38,6 +38,14 @@ public function getSalt(): ?string
         return null;
     }
 
+    public function getUserIdentifier(): string
+    {
+        return $this->username;
+    }
+
+    /**
+     * @deprecated use getUserIdentifier instead
```
DonCallisto, line 47: Is this fine?

DonCallisto:
Last deprecation detected in tests (https://github.com/knpuniversity/oauth2-client-bundle/runs/3808912774) is about the old authenticator, so is perfectly fine as long as older symfony versions are supported.

weaverryan:
Thank you Samuele!

PR #336, opened 2021-10-11
Add PHP 8.1 to GitHub actions

DonCallisto:
As PHP 8.1 will be out shortly and since this bundle claims to support a vast range oh PHP versions, I think it's good to add a build for PHP 8.1 also.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,7 @@ jobs:
                     - '7.3'
                     - '7.4'
                     - '8.0'
+                    - '8.1'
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks @DonCallisto! About 7.1, we could remove that, but if it doesn't cause any issues, I'm not in a rush. In a perfect world, I like to have our libraries support the PHP version of the Symfony LTS. Right now that is 4.4 and PHP 7.1, though 5.4 will be released in November, and then we can at least jump to 7.2. But that's not a strict rule: as soon as old versions cause any real complication, we can jump beyond them.

PR #345, opened 2021-12-09
Precision about the value of proxy in the settings

### LINKED ISSUE #344: Documentation is not accurate about http_client proxy settings
Took me hours to figure out that this configuration:

```
knpu_oauth2_client:
    http_client_options:
        proxy: null
        timeout: 5
```

has not effect on the `Guzzlehttp` client; value of `proxy` should be set to `false`, not `null`.

thejoelinux:
Fixes #344

### DIFF
--- a/README.md
+++ b/README.md
@@ -603,7 +603,7 @@ knpu_oauth2_client:
     # options to configure the default http client
     # http_client_options:
     #     timeout: 0
-    # if you want to disable the proxy (local gitlab oauth for example) set proxy to false
+    # if you want to disable the proxy (local GitLab OAuth for example) - set it to "false"
     #     proxy: null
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -603,6 +603,7 @@ knpu_oauth2_client:
     # options to configure the default http client
     # http_client_options:
     #     timeout: 0
+    # if you want to disable the proxy (local gitlab oauth for example) set proxy to false
```
bocharsky-bw, line 606:
```suggestion
    # if you want to disable the proxy (local GitLab OAuth for example) - set it to "false"
```

bocharsky-bw (approved):
Thank you!

PR #352, opened 2022-02-04
Add link to Generic Provider

ucay:
Instead of scrolling to the generic provider segment, link will make it easier.


bocharsky-bw (approved):
It's really helpful, thanks!

Issue #348, opened 2021-12-24
Symfony 6 `Symfony\Component\Security\Guard\AbstractGuardAuthenticator` not found


Myks92:
Class `Symfony\Component\Security\Guard\AbstractGuardAuthenticator` not found. Symfony 6 Remove Guard Component.  More: [Symfony 5.3: Guard Component Deprecation](https://symfony.com/blog/new-in-symfony-5-3-guard-component-deprecation). This dependency is used in [SocialAuthenticator](https://github.com/knpuniversity/oauth2-client-bundle/blob/2c925911cca6ed16b0695000bf927d405815dc36/src/Security/Authenticator/SocialAuthenticator.php#L24)

thacac:
Hi same issue here with symfony 6

Myks92:
> Hi same issue here with symfony 6

I think that we are not the only ones with such a problem. 

It is very strange that the bundle has already released Symfony 6 support and has not removed all deprecated. 

I hope they will fix it soon.

thacac:
I fixed it by extending OAuth2Authenticator instead and then set the new authenticator interface
`public function authenticate(Request $request): Passport
    {
        $service = $request->get('service');
        $client = $this->_getOauthClient($service);
        $token = $this->fetchAccessToken($client);

        /** @var $user ResourceOwnerInterface */
        $user = $client->fetchUserFromToken($token)->toArray();

        if ('github' === $service) {
            $user = $this->getGithubMail($user, $token);
        }

        $user = $this->_usersRepository->findOrCreateFromOauth($user, $service);

        return new SelfValidatingPassport(
            new UserBadge($token, function () use ($user) {
                return $user;
            })
        );
    }`

some are homemade methods, but here your can see the idea...hope this helps

Myks92:
Да, это отличная идея. Не хотите заниматься пиаром?

ToshY:
As expressed in [#316](https://github.com/knpuniversity/oauth2-client-bundle/issues/316#issuecomment-934889518) , you can use the [OAuth2Authenticator](https://github.com/knpuniversity/oauth2-client-bundle#using-the-new-symfony-authenticator) instead.

PR #353, opened 2022-02-04
Replace Guard with Symfony Custom Authenticator

bocharsky-bw:
@weaverryan Does Symfony Custom Authenticator makes more sense than Guard now?  I suggest moving this upper in order to show it _before_ the Guard one that is legacy now.


weaverryan:
Yes, this has caused some confusion - like #348. Thanks for this!

PR #354, opened 2022-02-04
Remove Requirements section completely

bocharsky-bw:
In order to match: https://github.com/knpuniversity/oauth2-client-bundle/blob/24baf0d2e62a732be52eacc445eb50ab65383107/composer.json#L16


weaverryan:
Actually, could we remove these? I've never found requirements helpful and they get out of date. WDYT?

bocharsky-bw:
Yeah, agree... anyway this info can be found in composer.json, and _there_ it's always fresh :)

PR #356, opened 2022-02-04
Add a link to the full list of clients from PHP League

bocharsky-bw:
It would be good to ask to double-check that official list first if there's no provider users are looking for in our table.


weaverryan:
Nice idea Victor!

PR #357, opened 2022-02-04
A bunch of good minor tweaks in the docs

bocharsky-bw:


README.md
```diff
@@ -46,15 +46,15 @@ Awesome! Now, you'll want to configure a client.
...
+### Step 1) Download the Client Library
 
 Choose the one you want from this list and install it
 via Composer:
 
 <a name="client-downloader-table"></a>
 
 | OAuth2 Provider                                                        | Install                                             |
-| ---------------------------------------------------------------------- | ------------------------------------------------------- |
+| ---------------------------------------------------------------------- | --------------------------------------------------- |
```
weaverryan, line 57: If this is too long, I think it needs to be fixed in `bin/update_readme`

bocharsky-bw: Ah, really! Now with #362 we won't forget about it :)

bocharsky-bw: The actual change is done in #360

PR #360, opened 2022-02-05
Make bin/update_readme script more dynamic and trim redundant spaces

bocharsky-bw:
The new version of the script is more dynamic now + a few minor tweaks, like trimming redundant spaces in the table and fixing some line indentation


weaverryan:
Thanks Victor! This was an old ugly script of mine - glad you were able to improve it!

PR #361, opened 2022-02-06
Twitch helix support

ErdemUyanik:
Added support for the vertisan/oauth2-twitch-helix bundle

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ via Composer:
 | [SymfonyConnect](https://github.com/qdequippe/oauth2-symfony-connect) | composer require qdequippe/oauth2-symfony-connect  |
 | [Strava](https://github.com/Edwin-Luijten/oauth2-strava)              | composer require edwin-luijten/oauth2-strava       |
 | [Stripe](https://github.com/adam-paterson/oauth2-stripe)              | composer require adam-paterson/oauth2-stripe       |
-| [Twitch](https://github.com/tpavlek/oauth2-twitch)                    | composer require depotwarehouse/oauth2-twitch      |
-| [Twitch](https://github.com/vertisan/oauth2-twitch-helix)             | composer require vertisan/oauth2-twitch-helix      |
+| [Twitch (deprecated)](https://github.com/tpavlek/oauth2-twitch)       | composer require depotwarehouse/oauth2-twitch      |
+| [Twitch Helix](https://github.com/vertisan/oauth2-twitch-helix)       | composer require vertisan/oauth2-twitch-helix      |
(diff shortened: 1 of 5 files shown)


src/DependencyInjection/Providers/TwitchHelixProviderConfigurator.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public function getLibraryHomepage()
+    {
+        return 'https://github.com/vertisan/oauth2-twitch-helix';
+    }
+
+    public function getProviderDisplayName()
+    {
+        return 'Twitch';
```
bocharsky-bw, line 49: Hm, probably `Twitch Helix` to be different from `TwitchProviderConfigurator`'s name?

README.md
```diff
@@ -1407,6 +1408,8 @@ knpu_oauth2_client:
         # will create service: "knpu.oauth2.client.twitch"
         # an instance of: KnpU\OAuth2ClientBundle\Client\Provider\TwitchClient
         # composer require depotwarehouse/oauth2-twitch
+        # This is deprecated and won't be supported by Twitch in the near future (Feb. 28 2022)
+        # use the twitch_helix one (directly below) if you create a new app
```
bocharsky-bw, line 1412: This is auto-generated section. Please, run `bin/update_readme` instead to update this correctly.

Also, if the old Twitch way is deprecated, please, update its name in `TwitchProviderConfigurator::getProviderDisplayName()` to `Twitch (deprecated)`

bocharsky-bw (commented):
Cool, thanks for digging on this deeper!

Also, please, take a look at failed builds

src/DependencyInjection/Providers/TwitchHelixProviderConfigurator.php
```diff
@@ -0,0 +1,56 @@
...
+class TwitchHelixProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        // no custom options
+    }
+
+    public function getProviderClass(array $config)
+    {
+        return 'Vertisan\OAuth2\Client\Provider\TwitchHelix';
```
weaverryan, line 26: Can you add `use` statement for this and below in `getClientClass()`? Then use `TwitchHelix::class`.

This is not done this way in most providers, simply because, iirc, they predate PHP 5.5 when this was added :)

bocharsky-bw:
This one is ready for review, I applied changes and fixed tests here

bocharsky-bw (approved):
@ErdemUyanik Thank you for starting working on it! This looks good to me now.

If you could give this new Twitch Helix with the latest changes a try - it would be great :)

jrushlow (approved):
Awesome work @ErdemUyanik && @bocharsky-bw

PR #289, opened 2021-03-04
Update work GitHub Actions

Yozhef:

PR #362, opened 2022-02-07
Bring back bin/update-readme status check to CI again

bocharsky-bw:
This feature was lost in #289 after migration from Travis to GitHub Actions


bocharsky-bw:
Now if `bin/update_readme` command will leave repo in a dirty state - Static CI will fail, see this as an example: https://github.com/knpuniversity/oauth2-client-bundle/runs/5089847177?check_suite_focus=true

sadikoff (approved):
hah so easy change =) 🚀

weaverryan:
Woohoo! Thank you Victor!

PR #369, opened 2022-05-21
Fix an error message

fabpot:
The error message currently lists the class names, but the configuration needs the keys.


jrushlow (approved):
Thank you @fabpot!

PR #372, opened 2022-07-24
[docs]: Add entrypoint functionality to Authenticator docs

robmeijer:
As best I can tell, the `start()` method is still required to automatically redirect to the "login" page when accessing a protected resource. This PR copies its use from the Guard Authenticator to the Symfony Authenticator section.


jrushlow (approved):
Makes sense to me - Thanks @robmeijer!

PR #379, opened 2022-10-01
Add certificate support for Azure provider

uncaught:
Adds certificate support to the **Azure provider**:

- Makes configuration option `client_secret` optional
- Adds configuration options `client_certificate_private_key` and `client_certificate_thumbprint`

See https://github.com/TheNetworg/oauth2-azure/pull/170 (merged) for the respective PR on the provider.

That PR is not released, yet, but there are no compatibility issues with this PR here. They can be installed independently without drawbacks.

Only to actually use certificates you had to have both PRs merged/released/installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -694,7 +694,11 @@ knpu_oauth2_client:
             type: azure
             # add and set these environment variables in your .env files
             client_id: '%env(OAUTH_AZURE_CLIENT_ID)%'
+            # client_secret is optional if you use a client certificate
             client_secret: '%env(OAUTH_AZURE_CLIENT_SECRET)%'
+            # Using a client certificate requires thenetworg/oauth2-azure > 2.1.1:
+            # client_certificate_private_key: '%env(OAUTH_AZURE_CLIENT_CERTIFICATE_PRIVATE_KEY)%'
+            # client_certificate_thumbprint: '%env(OAUTH_AZURE_CLIENT_CERTIFICATE_THUMBPRINT)%'
             # a route name you'll create
             redirect_route: connect_azure_check
             redirect_params: {}
--- a/README.md
+++ b/README.md
@@ -694,14 +694,15 @@ knpu_oauth2_client:
             type: azure
             # add and set these environment variables in your .env files
             client_id: '%env(OAUTH_AZURE_CLIENT_ID)%'
-            # client_secret is optional if you use a client certificate
-            client_secret: '%env(OAUTH_AZURE_CLIENT_SECRET)%'
-            # Using a client certificate requires thenetworg/oauth2-azure > 2.1.1:
-            # client_certificate_private_key: '%env(OAUTH_AZURE_CLIENT_CERTIFICATE_PRIVATE_KEY)%'
-            # client_certificate_thumbprint: '%env(OAUTH_AZURE_CLIENT_CERTIFICATE_THUMBPRINT)%'
             # a route name you'll create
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -694,7 +694,11 @@ knpu_oauth2_client:
             type: azure
             # add and set these environment variables in your .env files
             client_id: '%env(OAUTH_AZURE_CLIENT_ID)%'
+            # client_secret is optional if you use a client certificate
             client_secret: '%env(OAUTH_AZURE_CLIENT_SECRET)%'
+            # Using a client certificate requires thenetworg/oauth2-azure > 2.1.1:
+            # client_certificate_private_key: '%env(OAUTH_AZURE_CLIENT_CERTIFICATE_PRIVATE_KEY)%'
```
eerison, line 700: for custom parameters I would suggest implement some `client_extra_params`, because every provider could have some custom implementation.

eerison: Or maybe some option to pass a client configured!

uncaught: Why is this different from all the other custom parameters

eerison: IMO we we are just growing and growing this config!

for example

provider A, B, C, D ...: need the config
 - X1
 - X2
 - X3
 
 provider W: need the config
 - X4
 - X5
 - X6
 
  provider Y: need the config
 - X7
 - X8
 - X9
 
 
 most of the providers needs only the configs X1,2,3 But because few providers I need to add more and more config,
 I don't see it as a good approach!

src/DependencyInjection/Providers/AzureProviderConfigurator.php
```diff
@@ -12,11 +12,32 @@
 
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 
-class AzureProviderConfigurator implements ProviderConfiguratorInterface
+class AzureProviderConfigurator implements ProviderConfiguratorInterface, ProviderWithoutClientSecretConfiguratorInterface
```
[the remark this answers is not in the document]
uncaught: It's what the [AppleProviderConfigurator](https://github.com/knpuniversity/oauth2-client-bundle/blob/master/src/DependencyInjection/Providers/AppleProviderConfigurator.php#L15) uses, it's part of this repo:

https://github.com/knpuniversity/oauth2-client-bundle/blob/master/src/DependencyInjection/Providers/ProviderWithoutClientSecretConfiguratorInterface.php.

weaverryan:
This looks good to me - except two little code styling things to run:

A) `php-cs-fixer fix --verbose --diff` (once you have `php-cs-fixer` installed - you could even temporarily `composer require` it locally into this repo then just not commit the composer changes)

B) `bin/update_readme` - to regenerate part of the README

Cheers!

weaverryan (approved):

weaverryan:
Thanks @uncaught!

uncaught:
All green :) Is there something else missing?

weaverryan:
Thanks for the poke - this DID get merged in with a rebase... and for some reason my system didn't close the PR. It's all good and in!

uncaught:
> Hello @weaverryan , I am really interested in this feature. Do you know when you are going to release?
> 
> Thanks 🙏

Bare in mind, that https://github.com/TheNetworg/oauth2-azure/pull/170 hasn't been released, yet, either. So you could install that repository from master or had to wait for their release, too.

PR #381, opened 2022-10-31
Allow null as redirect_route

toooni:
The functionality “Fetching access keys via OAuth2 to be used with an API” mentioned in the README does not necessarily need a `redirect_url`, thus providing it should not be mandatory. We have an Endpoint (pingen.com) which does not allow `redirect_url` to be sent in these cases. Also, it wouldn't make much sense if it did because the route must be a valid symfony route, which in some cases must be created without a use.
I also think it makes sense to allow `null` here because this parameter isn't mandatory in the `League\OAuth2\Client\ProviderAbstractProvider` (https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L454).

### DIFF
--- a/src/DependencyInjection/ProviderFactory.php
+++ b/src/DependencyInjection/ProviderFactory.php
@@ -38,12 +38,14 @@ class ProviderFactory
      *
      * @return mixed
      */
-    public function createProvider($class, array $options, $redirectUri, array $redirectParams = [], array $collaborators = [])
+    public function createProvider($class, array $options, string $redirectUri = null, array $redirectParams = [], array $collaborators = [])
     {
-        $redirectUri = $this->generator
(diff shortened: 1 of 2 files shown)


weaverryan (approved):

weaverryan:
Thanks Toni - this makes sense :)

toooni:
@weaverryan Thank you for the fast merge 👍 Just fyi: I did not change anything about the config.  `redirect_url` is still required there, but using `redirect_url: null` will work. I am not sure if redirect_url should have a default value `null` since this bundle is mostly used for purposes which require a `redirect_url`.

weaverryan:
> I am not sure if redirect_url should have a default value null since this bundle is mostly used for purposes which require a redirect_route.

I agree. For the moment, I think we should keep the config required so that it's not "missed".

PR #389, opened 2023-02-01
Drop legacy PHP support

bocharsky-bw:
Apply some Rector rules to upgrade syntax up to PHP 7.4 and drop legacy PHP support below 7.4


weaverryan (approved):
Cool - well done - I like the updates with Rector!

weaverryan:
Thanks Victor!

PR #382, opened 2022-12-16
Added support to specify keycloak version

micbis:

PR #391, opened 2023-02-02
KeycloakProviderConfigurator.getProviderOptions() : add version to returned options array

pubphanguyen:
KeycloakProviderConfigurator.getProviderOptions() : add version to returned options array


src/DependencyInjection/Providers/KeycloakProviderConfigurator.php
```diff
@@ -57,6 +57,7 @@ public function getProviderOptions(array $config)
             'clientSecret' => $config['client_secret'],
             'authServerUrl' => $config['auth_server_url'],
             'realm' => $config['realm'],
+            'version' => $config['version'],
```
weaverryan, line 60: It does seem to make sense - the `version` option was added in #382, but it was never exposed here.

pubphanguyen:
Thank you @weaverryan for the merge!

PR #393, opened 2023-02-03
Set Keycloak provider's 'version' config option to null by default

### LINKED ISSUE #392: Keycloak optional config "version" raise cache error if not provided
Hey there, 

Since a recent update, I noticed that a "version" configuration is now added for Keycloak.
As it if qualified as "optional" in your documentation, omitting it raise an error while clearing the cache :

`In KeycloakProviderConfigurator.php line 60:

  Warning: Undefined array key "version"
`

bocharsky-bw:
Fixes #392


weaverryan:
Thanks Victor!

PR #395, opened 2023-02-17
feat: add Webflow provider

EmilePerron:
This PR adds an implmentation for the new [`koalati/oauth2-webflow`](https://github.com/koalatiapp/oauth2-webflow) OAuth 2.0 provider for [Webflow](https://webflow.com/).

### DIFF
--- a/src/DependencyInjection/Providers/WebflowProviderConfigurator.php
+++ b/src/DependencyInjection/Providers/WebflowProviderConfigurator.php
@@ -11,6 +11,7 @@
 namespace KnpU\OAuth2ClientBundle\DependencyInjection\Providers;
 
 use KnpU\OAuth2ClientBundle\Client\Provider\WebflowClient;
+use Koalati\OAuth2\Client\Provider\Webflow;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 
 class WebflowProviderConfigurator implements ProviderConfiguratorInterface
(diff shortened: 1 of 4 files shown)


EmilePerron:
Hi folks at @knpuniversity 👋

Is there anything blocking this PR from being merged in its current state?
If so, do let me know: I'll do any changes necessary to get this merged ASAP :)

src/DependencyInjection/Providers/WebflowProviderConfigurator.php
```diff
@@ -0,0 +1,55 @@
...
+class WebflowProviderConfigurator implements ProviderConfiguratorInterface
+{
+    public function buildConfiguration(NodeBuilder $node)
+    {
+        // no custom options
+    }
+
+    public function getProviderClass(array $config)
+    {
+        return 'Koalati\OAuth2\Client\Provider\Webflow';
```
bocharsky-bw, line 25: We can use `return Webflow::class` here too instead of hardcoding this as a string.. but don't forget to use the namespace too :)

bocharsky-bw (commented):
Hey, sorry for the delay! This looks good to me, only one minor comment

weaverryan (commented):
Yup - I agree with Victor - just that one change should do it :)

src/DependencyInjection/Providers/WebflowProviderConfigurator.php
EmilePerron: Done! :)

EmilePerron:
Thanks to the both of you - and no worries on the delay, you folks do more than enough!

I just changed the hardcoded string; the PHP 8.1 and 8.2 tests "failed" but the errors in these seemed to be deprecations in other parts of the codebase. 

Let me know if there's anything else you need me to do :)

Cheers!

bocharsky-bw (approved):
Thank you! It looks good me now

PR #398, opened 2023-03-21
Use remote GitHub CI config files

bocharsky-bw:
https://github.com/orgs/SymfonyCasts/projects/7/views/1?sortedBy%5Bdirection%5D=asc&sortedBy%5BcolumnId%5D=33394902&pane=issue&itemId=21066810

~We drop PHP 7.2 & 7.3 support here but we support those versions on other repos like VerifyEmailBundle and ResetPasswordBundle so that's why tests failed. I winder if we should drop 7.2/7.3 support there as well? Probably the easiest. Otherwise, we need a workaround for this in our global GitHub CI config files~

UPD it's possible to override matrix now


weaverryan:
I can see a lot of hard work here to get a nice, final version. Thanks!

PR #408, opened 2023-06-29
Rebuild CI and fix CS

bocharsky-bw:
After some changes upstream in GitHub CI config: https://github.com/SymfonyCasts/.github/pull/2

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,8 @@ jobs:
 
     cs:
         uses: SymfonyCasts/.github/.github/workflows/php-cs-fixer.yaml@main
+        with:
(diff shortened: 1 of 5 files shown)


bocharsky-bw:
Seems the lowest support version here is 7.4 so I had to downgrade the PHP version for PHP CS Fixer to 7.4 as well to avoid fixing:
```diff
-throw new UnsupportedUserException(sprintf('Instances of "%s" are not supported.', \get_class($user)));
+throw new UnsupportedUserException(sprintf('Instances of "%s" are not supported.', $user::class));
```

that is not supported by PHP 7.4

weaverryan:
Thanks Victor!

PR #425, opened 2023-11-28
Allow Symfony 7

bocharsky-bw:


composer.json
```diff
@@ -14,25 +14,25 @@
...
+        "symfony/routing": "^4.4|^5.0|^6.0|^7.0",
+        "symfony/http-foundation": "^4.4|^5.0|^6.0|^7.0",
         "league/oauth2-client": "^2.0"
     },
+    "require-dev": {
+        "league/oauth2-facebook": "^1.1|^2.0",
+        "symfony/phpunit-bridge": "^5.3.1|^6.0|^7.0",
+        "symfony/security-guard": "^4.4|^5.0|^6.0|^7.0",
+        "symfony/yaml": "^4.4|^5.0|^6.0|^7.0",
+        "phpstan/phpstan": "^0.12"
```
bocharsky-bw, line 28: We don't use this dep, can be removed

bocharsky-bw: Actually it's used

PR #431, opened 2024-02-12
hf: symfony 7

vitsadecky:
Fix `SaveAuthFailureMessage` trait for symfony 7

### DIFF
--- a/src/Security/Helper/SaveAuthFailureMessage.php
+++ b/src/Security/Helper/SaveAuthFailureMessage.php
@@ -14,6 +14,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
 use Symfony\Component\Security\Core\Security;
+use Symfony\Component\Security\Http\SecurityRequestAttributes;
 
 trait SaveAuthFailureMessage
 {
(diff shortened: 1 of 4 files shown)


src/Security/Helper/SaveAuthFailureMessage.php
```diff
@@ -22,7 +23,11 @@ protected function saveAuthenticationErrorToSession(Request $request, Authentica
         if (!$request->hasSession() || !$request->getSession() instanceof SessionInterface) {
             throw new \LogicException('In order to save an authentication error, you must have a session available.');
         }
+        
+        $securityClass = class_exists(SecurityRequestAttributes::class)
+            ? SecurityRequestAttributes::class
+            : Security::class;
 
-        $request->getSession()->set(Security::AUTHENTICATION_ERROR, $exception);
+        $request->getSession()->set($securityClass::AUTHENTICATION_ERROR, $exception);
```
bocharsky-bw, line 31: Hm, but `$securityClass` is a simple string right? I'm afraid we will have an error here. We need to set the error on a var above and use the var here

vitsadecky: it dosnet matter, its right this, maybe older php not 🤷 oki, i am goint to change it,  it will work, but its not pretty

vitsadecky: @bocharsky-bw its probably readable and more explicit, see agaign, thx 🙂

vitsadecky: i see pipeline failed, but its not related with this change 🤷

bocharsky-bw (approved):
Thank you!

Issue #434, opened 2024-02-20
Update documentation, drop unmaintained Symfony


tacman:
The documentation still refers to the "new" authentication system, how to implement the Guard authenticator, and the examples use the old annotations rather than attributes.

What do you think of a new release that requires PHP 8.1+ and Symfony 6.4+?  (I know, Symfony 5.4 is still maintained, but developers can use version 1 if they need that).

This is a great bundle, but the documentation seems to lean toward making it easy for people on old versions of Symfony rather than promoting best practices on newer ones.  For example, 

     enable_authenticator_manager: true

was only relevant from Symfony 5.1 to 6.2.  Authentication is already a bit complicated, I think the documentation should reflect usage based on current Symfony.  To refer to it as the "new" authentication manager when it was released nearly 4 years ago is confusing.

```bash
symfony new --webapp auth-demo && cd auth-demo
bin/console make:user
composer require knpuniversity/oauth2-client-bundle
```

Ideally, the developer could cut and paste from the README and it should work.

I'm happy to help with the documentation, as I often walk through it when I'm adding social media login to a new site.  But supporting  Symfony 4 at the expense of Symfony 6.4/7 seems wrong.

Related to #432

bocharsky-bw:
Hey @tacman ,

Thanks for raising this question! Here's my personal opinion on it:

The docs really need to be updated, agree. First of all, this bundle already supports PHP 8.1+,  see [composer.json](https://github.com/knpuniversity/oauth2-client-bundle/blob/6dbf6b6b92168da0265fd144f89ab34fb5a10f02/composer.json#L16) - and so let's drop annotations in favor of attributes in the docs.

About dropping Symfony 4.4 - let's drop its support already 👍  About 5.4 - I think it would be too aggressive. I would prefer to keep the 5.4 support, at least for one more year while it's actively supported by maintainers. This bundle has a simplified release policy, so it's easier to stick to  Symfony support versions.

> I think the documentation should reflect usage based on current Symfony

Yeah, I agree with it. We should promote instructions for the latest version, but also it would be handy to provide some comments and legacy code mentioning legacy versions. I'm against of dropping the legacy instructions completely from the docs as it may cause many issues for users on legacy versions.

If you have some ideas and could help with updating the docs - it would be much appreciated and warmly welcome! ❤️

tacman:
Can we at least move the Guard section to its own page, with a big ol' note that says "This is deprecated".  

[...]

PR #435, opened 2024-02-20
drop reference to .env.dist

tacman:
.env.dist was dropped in 2018.

https://symfony.com/blog/improvements-to-the-handling-of-env-files-for-all-symfony-versions

#434


jrushlow (approved):
Thank you @tacman!

Issue #447, opened 2024-07-20
Please update for Symfony > 7.1


Evgeny1973:
The **"Symfony\Component\HttpKernel\DependencyInjection\Extension"** class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from **"KnpU\OAuth2ClientBundle\DependencyInjection\KnpUOAuth2ClientExtension"**.

bocharsky-bw:
Any volunteers who want to help with this issue?

Evgeny1973:
Can't the authors of the bundle themselves add _\Extension_  to the end of the line _use Symfony\Component\DependencyInjection\Extension_??

PR #449, opened 2024-08-11
Fix Symfony 7.1 deprecation for Extension

Evgeny1973:
Fix deprecation ([issue 447](https://github.com/knpuniversity/oauth2-client-bundle/issues/447))
_The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "KnpU\OAuth2ClientBundle\DependencyInjection\KnpUOAuth2ClientExtension"._


bocharsky-bw (approved):
Thank you for taking the time to fix it! Looks like an easy win and tests are green

PR #456, opened 2025-05-12
Add return types to fetchAccessToken()

BenMorel:
Note that I did not use a native `: AccessToken` return type, as it would technically be a BC break (the class is not final and the method is not private)!


BenMorel:
@bocharsky-bw Can this be merged?

PR #464, opened 2025-10-17
Migrate XML config to PHP

dmaicher:
Fixes https://github.com/knpuniversity/oauth2-client-bundle/issues/461


phpstan.neon.dist
```diff
@@ -1,6 +1,8 @@
 parameters:
 	level: 5
 	inferPrivatePropertyTypeFromConstructor: true
+	bootstrapFiles:
+	    - ./vendor/symfony/dependency-injection/Loader/Configurator/functions.php
```
dmaicher: something is odd with the phpstan setup. This is normally not needed. Maybe its because its an outdated version (v2 is out since a while) or because its separate in the tools directory. Not sure.

Without this we get
image

dmaicher:
@bocharsky-bw shall we get this https://github.com/knpuniversity/oauth2-client-bundle/pull/465 merged then? I can finalize Symfony 8 support in https://github.com/knpuniversity/oauth2-client-bundle/pull/466 then

PR #465, opened 2025-10-19
Drop support for unmaintained Symfony versions

dmaicher:
I wanted to see how complex adding Symfony 8 support would be and noticed that we require `"symfony/security-guard": "^5.4"` which makes it hard to fully require Symfony 8 packages.

I'm proposing to drop support for unmaintained Symfony versions first. This will make it easier to add support for Symfony 8.

Note: this means removing the abstract `SocialAuthenticator` class. It cannot possibly work with Symfony > 5.4. I would not consider this a BC break because people on Symfony 5.4 will not get a new minor update anyway until they have to move to Symfony 6.4+ which means they need to change their authentication config anyway first. 

WDYT?


bocharsky-bw (approved):
Sf 5.4 is LTS and still supports security fixes, so I would usually keep maintaining it unless it complicates moving forward too much like in this case. So I agree to drop that, we still have 6.4 LTS, should be good enough IMO

PR #467, opened 2025-10-28
Update README for authenticator registration steps

necobm:
Removed instruction to enable authenticator manager in README since the option security.enable_authenticator_manager was deprecated since Symfony 6.2 and throws error in Symfony 7.3. 

See: https://github.com/scheb/2fa/issues/183

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,14 @@ security:
 +           custom_authenticators:
 +               - App\Security\MyFacebookAuthenticator
 ```
+> **IMPORTANT** If you have Symfony 6.4 or lower - you will also need to enable the new authenticator manager:
+
+```diff
+# app/config/packages/security.yaml
+security:
+    # ...  
++   enable_authenticator_manager: true
+```
 
 > **CAUTION** You *can* also inject the individual client (e.g. `FacebookClient`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -373,15 +373,13 @@ class MyFacebookAuthenticator extends OAuth2Authenticator implements Authenticat
...
+Next, register your authenticator
 in `security.yaml` under the `custom_authenticators` section:
 
 ```diff
 # app/config/packages/security.yaml
 security:
-    # ...
-+   enable_authenticator_manager: true
-  
+    # ...
```
bocharsky-bw: Hm, it could be too aggressive because this bundle still supports legacy Symfony versions `^5.4|^6.0|^7.0` and it may cause problems in legacy versions. How about keeping your changes, but also adding a note below this code block saying something like:

> If you have Symfony 6.4 or lower - you will also need to enable the new authenticator manager:
```diff
# app/config/packages/security.yaml
security:
    # ...
+   enable_authenticator_manager: true
```

Could you add this change too?

necobm: You have absolutely right! Added! Thanks for notice this

PR #473, opened 2026-03-19
allow client v3

### LINKED ISSUE #472: Please add support for league/oauth2-client 3

garak:
Fix #472


bocharsky-bw (commented):
Thank you for this PR!

I see we don't have v3 stable release yet, so CI indeed didn't test it on v3 yet

garak:
What if I change the last CI step (highest deps) to get "dev" as minimum stability?

bocharsky-bw:
I was thinking about doing it temporarily.. or maybe we're worrying about it too early? I mean, maybe we should just wait for a stable release?

PR #476, opened 2026-04-29
decrease schedule frequency

### LINKED ISSUE #475: Could we disable the schedule (or at least reduce it)?
Currently, there's a schedule configured in the CI actions that runs twice a day.
It seems largely exaggerated to me, and it wastes resources for the main repo and for every fork.
I'm not sure why this was introduced; I haven't seen any other repo doing the same.

Could we reconsider that choice, or at least put it on a reasonable schedule? Like every week, for example.

garak:
Fix #475

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,7 @@ on:
     push:
         branches: ['main']
     schedule:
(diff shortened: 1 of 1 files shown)


bocharsky-bw (approved):
Thank you!

